PR #202, opened 2011-06-17
Add hierarchical edge bundling layout.

jasondavies:
Based on http://cs.stanford.edu/people/mbostock/iv/dependency-tree.html.

A nicer way to specify `outgoing` links would be good.


hotwenlw:
Hi guys, Is it possible to differ the size of every stroke? 
How will it be done? Thanks for the help very much!

PR #204, opened 2011-06-19
Micro-optimisation: remove arguments check.

jasondavies:
This is around 6x faster when using V8's Crankshaft as the JIT bails due to the `arguments.length` reference.  I know the `arguments.length` check is used extensively in D3 and it's nice to keep it this way for consistency, but it might be worth replacing in this case as centroid may well get "hot" e.g. when used intensively for geo visualisations.

Still, micro-optimisations are debatable and this will likely become obsolete as Crankshaft gets smarter, so take with a pinch of salt!


akre54:
Might be worth revisiting this. Any function that leaks `arguments` or reassigns a var based on arguments is a [target for deoptimization](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#31-reassigning-a-defined-parameter-while-also-mentioning-arguments-in-the-body-typical-example) (no JIT'ing).

D3 does this all over the place so I'd be curious to see what perf improvements could be gained by a switch.

PR #277, opened 2011-08-28
Multi-value map support.

mbostock:
Current progress:
- selection.attr - done
- selection.style - done
- selection.property - done
- selection.classed - done
- selection.on - done
- transition.attr - done
- transition.style - done

Also, we may want to include these, but I'm not 100% yet:
- transition.attrTween?
- transition.styleTween?
- transition.tween?


coli:
I really wish this is supported..
```
selection.attr(function() { return {foo: 42}; });
```

mbostock:
@coli This is discussed in #2109, and I think we can do this for 4.0 as selection.attrs.

coli:
Thanks! 

Using each + select as a workaround just turns functional code back to being procedural code again.

PR #440, opened 2012-01-13
Make the voronoi implementation extensible.

bbroeksema:
This is a relative small change that makes d3.geom.voronoi more extensible. The current implementation assumes that users are only interested in the cell boundaries. However, for some code I'm working on I also need the neighbors of each cell.

In order to make this possible with as little changes as possible I propose this patch. In stead of calling d3.geom.voronoi only with a set of vertices it now also takes an optional callback parameter. This parameter should be an object that provides two methods:

* handle(e) // Handle edge addition
* result()    // finalize and return results.

In my patch the default behavior of d3.geom.voronoi is unchanged as I moved the original anonymous function into a default edgehandler which is used when the handler parameter is undefined.

This patch allows me to install my own edgehandler which records the required information.

Comments, questions and suggestions for improvements welcome of course.

Cheers,

Bertjan

### DIFF
--- a/src/geom/voronoi.js
+++ b/src/geom/voronoi.js
@@ -18,7 +18,7 @@
  * @param vertices [[x1, y1], [x2, y2], …]
  * @returns polygons [[[x1, y1], [x2, y2], …], …]
  */
-d3.geom.voronoi = function(vertices) {
+d3.geom.voronoi = function(vertices, edgeCallback) {
   var polygons = vertices.map(function() { return []; }),
       Z = 1e6;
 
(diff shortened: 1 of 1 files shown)


mbostock:
The new [d3-voronoi module](https://github.com/d3/d3-voronoi#voronoi-diagrams) for 4.0 returns the full Voronoi diagram, so it should be possible to track neighbors and other related operations. See the [Voronoi Topology example](https://bl.ocks.org/mbostock/cd52a201d7694eb9d890).

PR #517, opened 2012-02-20
Support HSV colors

alex:
This branch adds support for HSV colors. I implemented most of the API of HSL. What's not implemented (If it's desired), can easily be implemented by converting the HSV to HSL, doing the computation, and converting back.


mbostock:
With some help from @brewmook, I’ve created the [d3-hsv](https://github.com/d3/d3-hsv) module. Should be easy to implement d3.interpolateHsv and d3.interpolateHsvLong there, too.

PR #574, opened 2012-02-28
d3.layout.hierarchy: Allow internal nodes to have value

loganlinn:
This adds the ability for the internal nodes to contribute to their value, instead of their value being entirely composed of their children's values.

Adds a new method to d3.layout.hierarchy() to set `internal_value` in a similar fashion that you set the `value` for leaf nodes. When recursing, check to see if there's a function for internal_value, if there is, call it.

I am new to d3, so I apologize if this is already do-able, or if I left anything out.


mbostock:
This has been implemented in d3-hierarchy for 4.0!

PR #598, opened 2012-03-21
Ordinal scale inversion

jasondavies:
For use with d3.svg.brush e.g. for brushing histograms with ordinal scales.  This is a very simple implementation; ideally we want to take into account the current type of range in use.

If I'm headed in the right direction, I'd be happy to improve it and add tests. :)


SalahAdDin:
What's about this man?

curran:
This issue has been superceded by this other one https://github.com/d3/d3-scale/pull/60 There you can find an active effort towards this by @jheer and @mbostock .

It looks like ordinal scale inversion will land in D3 4.0.

curran:
Also here's an example that pretty much does invert an ordinal scale:

http://bl.ocks.org/mbostock/4349509

onursagir:
@curran inverting still seems absent, has this been implemented another way?

curran:
What do you mean? What would the semantics look like? What function is expected and what would be the input and output in particular?

PR #671, opened 2012-06-14
Force alpha tweaks

sj26:
Allow tweaking alpha decay rate and threshold.

### DIFF
--- a/src/layout/force.js
+++ b/src/layout/force.js
@@ -5,6 +5,8 @@ d3.layout.force = function() {
       size = [1, 1],
       drag,
       alpha,
+      alphaRate = 0.99,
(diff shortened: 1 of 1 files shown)


mbostock:
FYI, functionality similar to that proposed here is planned for D3 4.0 in [d3-force](https://github.com/d3/d3-force). I’m planning on exposing *simulation*.alphaDecay which sets the [exponential decay constant](https://en.wikipedia.org/wiki/Exponential_decay) such that the alpha for tick *i* is equal to *e*^(-*lambda* * *i*) where *lambda* is the decay constant. This is a bit more precise than multiplying the alpha value every iteration.

Sorry that it took me a few years! Related d3/d3-force#5.

PR #759, opened 2012-08-10
Add cubic polynomial splines

snoble:
Currently hermite splines are being used in a lot of places when you want to generate graphs of functions. Unfortunately these create bezier curves which may reverse eg
http://bl.ocks.org/3315490 and http://bl.ocks.org/3316746

This can be solved by choosing control points that create regular polynomial splines
see http://bl.ocks.org/3316815 and http://bl.ocks.org/3316859

Tests are broken here because the generated paths have changed


wayx:
Hello. Any updates with that bug?

tandu:
I have used suggested replacement

mbostock:
This has been fixed with the new d3.curveMonotoneX and d3.curveMonotoneY in [d3-shape](https://github.com/d3/d3-shape) for [D3 4.0](https://github.com/mbostock/d3/tree/4). Here’s an updated version of Jason’s test:

http://bl.ocks.org/mbostock/972e360a214dc778b6d1c6245af7dc12

The new d3-shape module also implements Yuksel *et al.*’s [parameterized Catmull–Rom curves](http://www.cemyuksel.com/research/catmullrom_param/), defaulting to centripetal splines, which do not form loops or cusps within a curve segment.

snoble:
Thanks @mbostock!

Issue #668, opened 2012-06-06
d3.mouse return incorrect position when browser zoomed.


frogcherry:
d3.mouse performs a little strange in mouse events. When the svg did not on the left of screen, if I use the zoom tool of browser, it can’t recognize the location of mouse correctly. It happens in IE9, chrome, safari, opera, only in firefox it works well.

I made a simple **demo** of that: http://walkpie.cnodejs.net/
You can review the demo and download the code in the demo page.

The core part of the demo:
```html
<table>
<tr>
<td width="35%" style="background-color: steelblue;">This is Left.</td>
<td><div id="container"></div></td>
</tr>
</table>
```

```javascript
var vis = d3.select("#container")
    .append("svg")
    .attr("width",600)
    .attr("height",400);

vis.append("rect")
    .attr("width",600)
    .attr("height", 400)
    .attr("style", "fill:none;stroke:red;");

var cl = vis.append("circle")
    .attr("cx", 50)
    .attr("cy", 50)
    .attr("r", 5);

vis.on("mousemove", function(){
    var cp = d3.mouse(this);
    cl.attr("cx", cp[0])
      .attr("cy", cp[1]);
});
```

mbostock:
As far as I can tell, this has been fixed in current browsers. However, it is related to d3/d3-selection#67.

PR #766, opened 2012-08-13
Fix for WebKit bad CTM when zoom != 1

vogievetsky:
Fixes these:
#668
https://groups.google.com/d/topic/d3-js/tqJ_V7cV7Eg/discussion

A good place to test to see the diff is the mouse example.
Also I ran make so it changed d3.v2.js and the min... not sure what the pull request equate is for those files.
All the 'real' changes are in core/mouse.js

I refactored your code a little bit so that there is a d3_mouse_getScreenCTM function that just returns the 'fixed' CTM.

I added some extra code to set the html and body position to 'static' before doing the bug test. This is so that the test is more robust - an insane person might have body or html not be at (0,0) and have position not static.

Example:
```javascript
d3.select('body')
  .style('position', 'absolute')
  .style('left', 0)
  .style('top', '10px');

window.scrollTo(0, 10);
```
Add this to your code and it would trick the regular d3.mouse function to think that bug44083 exists when it might not.


mbostock:
As far as I can tell, this has been fixed in current browsers. However, it is related to d3/d3-selection#67.

PR #786, opened 2012-08-24
Adding y1 to default stackOut

ZJONSSON:
SVG Rectangles start at top-left corner (x0,y1).  Providing y1 simplifies rendering and should be backwards compatable.


mbostock:
The new [d3.stack](https://github.com/d3/d3-shape#stacks) for 4.0. outputs [*y0*, *y1*].

PR #886, opened 2012-11-10
code coverage test

rgarcia:
Added a "test-cov" makefile target that runs [jscoverage](https://github.com/visionmedia/node-jscoverage) on /src prior to testing. Outputs a coverage.html file with a line-by-line breakdown of code coverage. Currently clocking in at 71% coverage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 _site
 node_modules
 .DS_Store
+src-cov
+coverage.html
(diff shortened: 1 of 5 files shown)


mbostock:
Sorry for not responding to this pull request. The repository structure is changing for D3 4.0 (see <https://github.com/d3>), so this work would need to be modified for the new repositories… but perhaps it’s a good idea. Perhaps using [Istanbul](https://github.com/gotwarlost/istanbul)?

PR #1055, opened 2013-02-05
Allow custom coordinate space while dragging

darwin:
Problem: when you modify grandparent's transform during drag callback you get discontinuity in d3.event data. This leads to poor experience during dragging as described here: http://stackoverflow.com/questions/13078535/stuttering-drag-when-using-d3-behavior-drag-and-transform

Solution: allow specification of custom node which will be taken as base coordinate space for the drag session.

In our case of modification grandparent's transform, this usage fixes the issue:

```
d3.behavior.drag().base(function(target) { return target.parentNode.parentNode; });
```

### DIFF
--- a/src/behavior/drag.js
+++ b/src/behavior/drag.js
@@ -1,6 +1,7 @@
 d3.behavior.drag = function() {
   var event = d3_eventDispatch(drag, "drag", "dragstart", "dragend"),
-      origin = null;
+      origin = null,
+      baseNode = function(target) { return target.parentNode; };
 
   function drag() {
     this.on("mousedown.drag", mousedown)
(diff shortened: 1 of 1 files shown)


ebengtso:
I'm having the same problem.

I have a rectangle with resize handles
```
<g>
<g id="rects" transform="translate(x,y)"><rect></g>
<g id="handles" transform="translate(x,y)"><rect id="handle1"><rect id="handle2"></g>
</g>
```
I have added a drag listener on the handle, and then I resize the rect group and apply a transform on both rects and handles.

Changing the position of the parent during drag reproduces the problem

thesamprice:
Ran into this same issue. D3 now has the container function that cleans up jitter.
```
        let dragGroup = d3.select('#parentContainerId')
        let drag = d3.drag()
        .on("start", this.PieceDragStarted.bind(this))
        .on("drag",  this.PieceDrag.bind(this))
        .on("end",   this.PieceDragEnd.bind(this))
        .container(dragGroup.node().parentNode)
         g.call(drag)

```

Issue #1084, opened 2013-02-19
Allow to restrict pan when zooming


dwt:
There is a zoom.scaleExtent() to only allow sensible values, but there is no panExtent() that allows me to preven the graph from scrolling out of the visible area.

That seems to be an oversight?

jasonbodily:
The feature's not part of the official d3 library at the moment. 

Also NOTE: If scale limits are less than 1 then the bounding functions offered by RByers and myself no longer work as expected. Working on a fix.

lebolo:
@pedroteixeira you can use PR #2019 as a start off point. There are merge conflicts again since it's been in queue for so long, but it may help you to roll your own solution.

mbostock:
This has been implemented as *zoom*.extent and *zoom*.translateExtent in [d3-zoom](https://github.com/d3/d3-zoom) for D3 4.0.

PR #1087, opened 2013-02-20
Attempt at zoom.translateExtent.

mbostock:
Doesn’t work correctly because the translateExtent may need to be dependent on
scale, or the zoom behavior may need to know the visible area. Related #1084.


mbostock:
This has been implemented as *zoom*.extent and *zoom*.translateExtent in [d3-zoom](https://github.com/d3/d3-zoom) for D3 4.0.

Issue #711, opened 2012-07-17
d3.time.day incorrectly handles dates earlier than year 100


bunkat:
It seems like d3.time.day (and possibly other time functions) treats any 2 digit year as being in the 1900's even if date.getFullYear() returns a value below 100.

For example, using d3.time.day on the date: 

```
Date {Wed Jul 09 0098 01:39:51 GMT-0700 (LMT)}
```

returns

```
Date {Thu Jul 09 1998 00:00:00 GMT-0700 (PDT)}
```

You can see an example of the problem at http://bl.ocks.org/3129952.  Move the selection rectangle on the bottom axis to the left until it crosses the year 100 threshold.  It will disappear and the top axis will start showing dates in the 1900's.  

Modifying

```
var rects, labels
    , minExtent = d3.time.day(brush.extent()[0])
    , maxExtent = d3.time.day(brush.extent()[1])
```

to

```
var rects, labels
    , minExtent = brush.extent()[0]
    , maxExtent = brush.extent()[1]
```

allows the chart to work as expected.

PR #1197, opened 2013-04-13
Fix Firefox time interval bug for Europe/London.

### LINKED ISSUE #1196: Firefox mangles date axis (possibly only during BST?)
Example from an old stackoverflow question:
http://jsfiddle.net/nrabinowitz/JTrnC/

In firefox the date axis is filled with "01 AM" markers instead of the dates visible in Chrome and IE.

Stepping through, it looks like Firefox is translating dates with no time as occuring at 1AM, leading d3_time_scaleFormat() to select the format that returns getUTCHours() rather than the day and date.

Would it be better to use getHours() as the user is probably more interested in local times (I know I'm opening a can of worms here)?

BST is UTC+1, by the way, what we laugingly refer to in the UK as British "Summer" Time.

jasondavies:
d3.time.day should always return midnights.  However, Firefox doesn't
return midnight for the following:

```
var d = new Date(1970, 0);
d.setFullYear(2011, 8, 22);
// Thu Sep 22 2011 01:00:00 GMT+0100 (BST)
```

Chrome and Node.js return:

```
// Thu Sep 22 2011 00:00:00 GMT+0100 (BST)
```

This is because `new Date(1970, 0)` has different daylight saving
behaviour in the two browsers.

Chrome: Thu Jan 01 1970 00:00:00 GMT+0000 (GMT)
Firefox: Thu Jan 01 1970 01:00:00 GMT+0100 (BST)

In fact, Firefox is more correct!  The UK was on permanent BST from
February 1968 until October 1971.  This seemingly minor difference in
interpretation results in non-midnight times being returned for all day
intervals and other intervals that use d3.time.day internally.

The fix forces midnight to always be returned.

Fixes #1196.


mbostock:
> Sorry for being dim but what's the simple fix you recommend for displaying the year and not 1am on the axis?

The other way to fix this is to explicitly specify an axis.tickFormat (using `d3.format("%Y")`, for example), rather than relying on the default multi-scale time format and dates that are precisely aligned with time interval boundaries.

mbostock:
Resurrecting a thread here because I’m starting work on the new d3-time-interval module…

It seems like the main cause of the pain here is (ironically) caused by an earlier fix 83b5f459979a062a1da38d4675be746788697883 for #711, which changed the day method to workaround the abysmal behavior of the Date constructor regarding two-digit years. Before it was:

```js
function day(date) {
  return new Date(date.getFullYear(), date.getMonth(), date.getDate());
}
```

And after the fix:

```js
function day(date) {
  var day = new Date(0, date.getMonth(), date.getDate());
  day.setFullYear(date.getFullYear());
  return day;
}
```

This introduced the new bug where the wrong date could be returned if the timezone changed between 1900 (and later, after 3847172c6f1199acfb94715064eaf027298c4958, 2000) and the current date’s year.

But, rather than using the search method in this pull request, wouldn’t it be sufficient to do something more like this? (I think you proposed this previously, but I can’t find it now.)

```js
function day(date) {
  var y = date.getFullYear(),
      m = date.getMonth(),
      d = date.getDate();
  if (0 <= y && y <= 99) {
    date = new Date(-1, m, d);
    date.setFullYear(y);
  } else {
    date = new Date(y, m, d);
  }
  return date;
}
```

This assumes that the year -1 (1 B.C.) has the same daylight savings behavior as years in the range [0,99], but I think that’s a safe assumption. And then we shouldn’t need to do anything fancier to floor the date.

(We still need to change the predicates to fix the non-midnight boundary issue, but this should at least return what the browser thinks is the correct definition of a day.)

mbostock:
Another solution?

```js
function day(date) {
  date = new Date(+date);
  date.setHours(0, 0, 0, 0);
  return date;
}
```

mbostock:
This code has been completely rewritten in [d3-time](https://github.com/d3/d3-time) for 4.0.

PR #1250, opened 2013-05-16
enable configuration of a different window object

sammyt:
Presently when running in node.js, d3 creates it's own window object to operate on.  While initially convenient this makes it quite difficult to have independent flows of execution all using d3.

This change enables the creation of new d3 instances which are bound to provided windows.

My driving motivation is that I want to use d3 to build my views on both the client and the server.  On the server I need each request to have an isolated environment.

Here is an example of using the proposed `d3.env` function with jsdom

```javascript
jsdom.env(
  "<html><head/><body><div class='wibble'></div></html>"
, function(err, window){
    var d3 = require('d3').env(window)
    d3.select('.wibble')
      .classed('wibble', false)
      .classed('wobble', true)
    console.log(window.document.innerHTML)
    // <html><head></head><body><div class="wobble"></div></body></html>
  }
)
```

As well as allowing for isolated windows this enables the use of alternative DOM implementations.  As an example for my own purposes I believe [domino](https://github.com/fgnass/domino) is more appropriate.

```javascript
var domino = require('domino')
  , win = domino.createWindow('<head></head><body><div id="foo"></div></body>')

var d3 = require('d3').env(win)
d3.select('#foo').classed('dimdom', true)
console.log(win.document.innerHTML)
// <html><head></head><body><div id="foo" class="dimdom"></div></body></html>
```

Grateful for you thoughts
sammy


mbostock:
Fixed in #2225; D3 no longer assumes a global window or document. So, if there’s no global window or document, you must call d3.select(node).select(string) to first specify the root node of a selection rather than calling d3.select(string) which tries to use the global document.documentElement.

PR #1262, opened 2013-05-21
Include an API for SVG transform attributes

trinary:
d3-transform makes it easy to define and reuse functions that produce transform attribute strings for SVG elements. Using d3-transform reduces repetition, allows you to compose multiple transforms, and eliminates ugly string-interpolation from transform attr calls.

More info and docs here: https://github.com/trinary/d3-transform#readme

This PR includes that transform API and adapts the code and tests in that repo to fit d3's existing structure and style.

### DIFF
--- a/src/svg/index.js
+++ b/src/svg/index.js
@@ -10,3 +10,4 @@ import "diagonal-radial";
 import "symbol";
 import "axis";
 import "brush";
+import "transform";
--- a/src/svg/transform.js
+++ b/src/svg/transform.js
@@ -1,0 +1,37 @@
+(function() {
+  d3.svg.transform = function(chain) {
+    var transforms = [];
+    if (chain !== undefined) { transforms.push(chain) }
+
+    function push(kind, args) {
+      var n = args.length;
+
+      transforms.push(function() {
+        return kind + '(' + (n == 1 && typeof args[0] == 'function'
+            ? args[0].apply(this, arr(arguments)) : args) + ')';
+      });
+    };
+
+    function arr(args) {
+      return Array.prototype.slice.call(args);
+    }
+
+    var my = function() {
+      var that = this,
+          args = arr(arguments);
+
+      return transforms.map(function(f) {
+        return f.apply(that, args);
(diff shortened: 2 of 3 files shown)


dexygen:
A year and a half later and this still hasn't been merged.  I'm a complete newbie to d3 and it's enough of an uphill climb without the error-prone string concatenations (to the point where I am tempted to put my arguments into an array and do the concatenation in one fell swoop with join)

mgold:
Turning an array to a string includes the comma but not the brackets. You still need `makeArray` but not `join`. Alternatively, throw this in your code:

````js
d3.selection.prototype.translate = function(a, b) {
  return arguments.length == 1
      ? this.attr("transform", "translate(" + a + ")")
      : this.attr("transform", "translate(" + a + "," + b + ")")
};
````

Now you can call `anySelection.translate(anX, aY)` or `anySelection.translate([anX, aY])` with method chaining. Caveats: Does not get the value with 0 arguments, does not accept a function, clobbers any other SVG transforms you may have. So it's definitely not PR-worthy but can help you out if you know what you're doing.

See also: [d3.transform](https://github.com/mbostock/d3/wiki/Math#2d-transforms), which returns a normal object and not the standard method-chainable closure, and [trinary/d3-transform](https://github.com/trinary/d3-transform), a third-party library which does exactly that, but is IMHO a little too heavyweight for just translate (hence my prototype addition). Arguably, trinary's library should replace the current d3.transform in 4.0.

**Comment by mbostock** [comment]
In 4.0, the d3.transform class becomes private, internal to [interpolateTransform](https://github.com/d3/d3-interpolate#interpolateTransform). It should be possible to make a d3-transform plugin (module) that exposes a convenience API for specifying SVG transforms.

Though, I still think ES6 string templates are nice, too:

```js
d3.select("svg").selectAll("g")
    .data([{size: 5}, {size: 10}])
  .enter().append("g")
    .attr("transform", function(d) { return `translate(20,{d.size * 10})rotate(40)scale({d.size + 2})`); });
```

PR #1299, opened 2013-06-08
Add xExtent and yExtent to d3.behavior.zoom (fixes #1084)

### LINKED ISSUE #1084: Allow to restrict pan when zooming

mhsmith:
These limit all actions to ensure the domain of the corresponding scale always stays within the given extent. As well as panning, zooming actions may also be limited in the zoom-out direction. If more than one of scaleExtent, xExtent and yExtent are provided, the effective bottom end of the scale extent will be the most restrictive of the three.

If scaleExtent and x/yExtent cannot be satisfied simultaneously (i.e. the lower scale limit implied by x/yExtent is above the scaleExtent), then x/yExtent will take priority.

Simple example: http://bl.ocks.org/mhsmith/5732011
Example with discontinuous and log scales: http://bl.ocks.org/mhsmith/5734904

### DIFF
--- a/src/behavior/zoom.js
+++ b/src/behavior/zoom.js
@@ -15,8 +15,10 @@ d3.behavior.zoom = function() {
       event = d3_eventDispatch(zoom, "zoom"),
       x0,
       x1,
+      xExtent = d3_behavior_scaleInfinity,
       y0,
       y1,
+      yExtent = d3_behavior_scaleInfinity,
       touchtime; // time of last touchstart (to detect double-tap)
 
   function zoom() {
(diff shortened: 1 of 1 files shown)


mbostock:
Merging into #2019.

mbostock:
This has been implemented as *zoom*.extent and *zoom*.translateExtent in [d3-zoom](https://github.com/d3/d3-zoom) for D3 4.0.

PR #1343, opened 2013-06-26
Implement inertial drag

pjanik:
Please see the example: http://bl.ocks.org/pjanik/5872514
Recently I needed such feature and implemented it in my application, but then I came up with the idea that it can be easily generalized. Inertial drag is quite popular nowadays.

I'm creating pull request so you can easily review the code, but of course I'm open for discussion. 
Do you find it useful?

Note that I'm not very familiar with d3 codebase, so I'm happy to fix implementation details, style, add some tests etc. in case of need.


vanquang9387:
I really need this feature. Will this be integrated?

langdonx:
@mbostock No chance for this or is there an existing solution?

mbostock:
I bulk-closed all pull requests in this repository because the code was moved to separate repositories as part of the D3 4.0 launch (and thus no pull requests here could be merged). See d3/d3-drag#27 and d3/d3-zoom#67 for the new issues.

PR #1354, opened 2013-06-30
Accept function for selection append and insert.

mbostock:
Like selection.select, selection.append and selection.insert can now accept a function which returns a node. This facilitates appending or inserting elements whose name is computed from data, or appending elements that were created previously (such as removed elements or elements from a pool).

There has been [much discussion](https://github.com/mbostock/d3/pull/732#issuecomment-7436280) regarding whether the function should return the name of the element or the element itself. Returning a name is less work for the caller, but only supports creating new elements. Returning a name is more consistent with how D3 defines attribute values, but D3 does not allow attribute names to be specified as functions. So, it seemed better to opt for consistency with selection.select and selection.selectAll, which accept functions that return elements. Of course, you can still pass a function to selection.select, selection.selectAll or selection.each to accomplish the same thing, but using selection.append is more intuitive.

Related #4 #311 #724 #732 #734 #961 #1031 #1271.

### DIFF
--- a/test/selection/append-test.js
+++ b/test/selection/append-test.js
@@ -25,6 +25,12 @@ suite.addBatch({
         assert.isTrue(svg[0][0].parentNode === body.node());
         assert.isTrue(svg[0][0] === body.node().lastChild);
       },
+      "appends an element specified as a function": function(body) {
+        var svg = body.select("svg").remove().node();
+        assert.isFalse(svg === body.node().lastChild);
+        body.append(function() { return svg; });
+        assert.isTrue(svg === body.node().lastChild);
+      },
       "propagates data to new element": function(body) {
         var data = new Object(), div = body.data([data]).append("div");
         assert.strictEqual(div[0][0].__data__, data);


tracycollins:
Hello,

I found this thread while trying find a solution to avoid javascript garbage collection on a dynamic force directed graph. I have a graph where nodes "age out" and are removed from the graph, and new nodes are created dynamically from streaming data.

I think using a pool of DOM objects (i.e., pool of circles) that are reused on append and recycled on remove is the way to go.  To implement this, I think I'd need to use a function in the append and remove statements that pull elements from the pool on append and recycle to the pool on remove, right?

If anyone can point me to implementation examples, that would be great.

Thanks

p.s. Awesome library!

curran:
@tracycollins I think what you need here is the General Update Pattern, taking advantage of the `key` function to specify which DOM nodes should associate with which data elements.

If you have a constant number of nodes all the time, it should be possible to keep track of the `key` values of the graph nodes that get removed, then use those same values for the new graph nodes. This way, the DOM element for the aged out graph node will be immediately reused for the new graph node, and the DOM element will never get garbage collected.

PR #1557, opened 2013-10-02
Optimized array push() used in many files

Yahasana:
X-Ref 1bec3a67b


mbostock:
This optimization is largely adopted in the new [D3 modules](https://github.com/d3). Since the code has been almost completely rewritten (and moved to new repositories!), there’s no way to merge this pull request. But if you find any more slow code in the new modules please let me know!

PR #1652, opened 2013-11-29
Allow multiple locales.

### LINKED ISSUE #750: Format strings should be overridable.
d3.time.format contains a number of locale-specific strings (month names and day names, also the locale default formats for %x and %X), as well as regular expressions derived from these names. It would be nice to be able to customize these for a different locale.

d3.format also would benefit from localization. For example, you might want to format 1234.56 as "1 234,56" in France and  "1.234,56" in Germany. It’s possible to change the thousands group separator (currently a comma, ","), but the period comes from JavaScript’s built-in number formatting. So this might only be possible by substituting a custom number format implementation or by post processing to swap characters (e.g., "," ↦ " " and "." ↦ ","). d3.format does not currently support parsing so this simplifies the problem somewhat.

Doing this “properly” might be out-of-scope of D3; see for example [jquery/globalize](https://github.com/jquery/globalize), which is a much more extensive library for localization.

mbostock:
Fixes #750. Code that requires localization is now within a closure that has access to locale-specific strings.

To minimize the degree of isolation required, d3.scale.linear (and other quantitative scales, including d3.time.scale) still default to the en-US locale. However, you can create your own localized formats using locale.numberFormat and locale.timeFormat.

(It would be possible to move these other classes to within d3.locale as well, but that would mean anything that depends on d3.scale.linear would also need to be moved within d3.locale, such as d3.svg.axis. I think it’s cleaner to leave them using a default locale rather and have people localize the formats explicitly, rather than classes that localize anything downstream; you can always reassign d3.format to a different locale if you want to change the default.)

Also introduces locale.timeFormat.multi which can be used to create a [multi-scale time format](http://bl.ocks.org/mbostock/4149176), and improves currency formatting by allowing the currency symbol to appear after the numeric value.

### DIFF
--- a/src/locale/ru-RU.js
+++ b/src/locale/ru-RU.js
@@ -1,6 +1,6 @@
 import "locale";
 
-var d3_locale_ruRU = d3.locale({
+d3.locale.ru_RU = d3.locale({
   decimal: ",",
   thousands: " ",
   grouping: [3, 3],
--- a/test/locale/locale-test.js
+++ b/test/locale/locale-test.js
@@ -1,0 +1,116 @@
+var vows = require("vows"),
+    load = require("../load"),
+    assert = require("../assert"),
+    time = require("../time/time"),
+    local = time.local;
+
+var suite = vows.describe("d3.locale");
+
+suite.addBatch({
+  "locale": {
+    topic: load("locale/ru-RU").expression("d3.locale.ru_RU"),
+
+    "numberFormat": {
+      topic: function(locale) {
+        return locale.numberFormat;
+      },
+      "formats numbers": function(format) {
+        var f = format(",.2f");
+        assert.equal(f(12345.67), "12 345,67");
+      },
+      "formats currencies": function(format) {
+        var f = format("$,.2f");
+        assert.equal(f(12345.67), "12 345,67 руб.");
+      }
+    },
+
+    "timeFormat": {
+      topic: function(locale) {
+        return locale.timeFormat;
+      },
+
+      "format": {
+        "formats locale date and time": function(format) {
+          var f = format("%c");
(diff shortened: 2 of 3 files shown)


Monduiz:
I'm tryin to wrap my head around this. Does anyone have an example, say, on how to apply d3.locale to a bar chart to change dot-separated decimals to comma? This would go a long way to help understand how to use it in the most common scenarios.

wahlatlas:
Here's how it works

e.g. for spanish

var ES = d3.locale({
decimal: ",",
thousands: ".",
grouping: [3],
currency: ["", " €"],
dateTime: "%A, %e de %B de %Y, %X",
date: "%d/%m/%Y",
time: "%H:%M:%S",
periods: ["AM", "PM"],
days: ["domingo", "lunes", "martes", "miércoles", "jueves", "viernes",
"sábado"],
shortDays: ["dom", "lun", "mar", "mié", "jue", "vie", "sáb"],
months: ["enero", "febrero", "marzo", "abril", "mayo", "junio",
"julio", "agosto", "septiembre", "octubre", "noviembre", "diciembre"],
shortMonths: ["ene", "feb", "mar", "abr", "may", "jun", "jul", "ago",
"sep", "oct", "nov", "dic"]
});

    var thsd = ES.numberFormat("n");
    thsd(10000);  // 10.000

see here for options
https://github.com/mbostock/d3/wiki/Localization

see it in action here
https://www.destatis.de/bevoelkerungspyramide/

Monduiz:
Thank you, wahlatlas! It works beautifully. I implemented it today in my app and it saved me a lot of trouble. Thank you Mike for this wonderful functionality. I use more and more D3 now.

PR #1799, opened 2014-03-20
Pointer events for the drag behavior.

mbostock:


natevw:
Any particular feedback why this was "iceboxed"? I need to review more fully — and figure out whether this patch still applies cleanly — but it's looking like a feature in an app I maintain may have relied on some of these fixes (and of course for IE support) after I switched to a different custom-patched D3 build that neglected this work. Ideally it would all just be in trunk at some point if possible.

mbostock:
Lack of time, at the moment.

natevw:
xref as I pick this back up: http://bl.ocks.org/mbostock/9631744 is the nice little playground used for trying out d3.drag tweaks last round.

natevw:
This patch definitely does not rebase cleanly, and most of its multitouch-related fixes were applied separately in the meantime.

I've re-done the Win8 (i.e. Pointer Events) work as https://github.com/mbostock/d3/pull/2403 — @mbostock perhaps you could close this and review that instead?

mbostock:
Related #2556.

mbostock:
Superseded by #2403.

PR #1819, opened 2014-04-02
Add voronoi.topology.

mbostock:


sbryfcz:
Is there any update on this? I'd love to be able to run this awesome functionality on a node backend. It seems that the old build that contained this functionality is not node compatible. I get lots of errors. Document not being defined and such. Any suggestions for how I could proceed? I love how D3 and topojson has become very rich with functionality for manipulating topology. Keep up the great work!

mbostock:
I’m thinking about the new API over in the [d3-voronoi module](https://github.com/d3/d3-voronoi); see d3/d3-voronoi#1 and d3/d3-voronoi#2.

IftachSadeh:
Hey Mike,

Thanks for the awesome work on this project!
The  `voronoi.topology()` functionality is great. I see it's also not yet part of ` d3-voronoi`. Is there any way for one to use it with the current master version of d3?

Thanks!

mbostock:
Although I opted not to build this functionality into d3-voronoi directly, the latest release 0.3 of d3-voronoi makes the underlying Voronoi diagram available, so it’s now possible to do this in “user land” without touching the d3-voronoi implementation. For example:

[cd52a201d7694eb9d890](http://bl.ocks.org/mbostock/cd52a201d7694eb9d890)

If there’s demand, I’ll consider adding this to d3-voronoi, but I think it makes sense as an optional extension to d3-voronoi at the moment.

Issue #894, opened 2012-11-13
Removing behaviors


tmcw:
While it's possible to specifically unbind a behavior's handlers, I'm not seeing any clear way to easily remove the behavior itself, in v2 or v3 - if you want an element to be draggable most of the time, but disable the drag handler later.

PR #1855, opened 2014-04-23
Remove all listeners with selection.on("*", null).

jasondavies:
This removes all listeners (namespaced or not) for all elements in a given selection.

Related discussion in #894.  I haven’t personally needed to use this but we can discuss further in this pull request whether it’s worth supporting.


burner986:
I like that here is a lot of discussion about how to remove listeners. In my case after I remove a listener, i want to later bring it back. How could I achieve that? Using google I came upon this lines: 
var dragCallback = d3.select('rect#no-drag').property('__onmousedown.drag')['_'‌​];
d3.selectAll('rect#no-drag').on('mousedown.drag', dragCallback);
But it doesn't seem to work anymore in d3 v4. I have an open stackoverflow question for this too: http://stackoverflow.com/questions/42245111/dynamically-resize-a-div-when-it-is-used-as-a-node-in-a-d3-force-directed-graph

mbostock:
@burner986 Please use Stack Overflow to ask for help, and don’t cross-post here. You can retrieve a named event listener using *selection*.on like so:

```js
var listener = selection.on("mousedown.drag");
```

So, here’s how you could remove it, and then add it back:

```js
var listener = selection.on("mousedown.drag");
selection.on("mousedown.drag", null); // removed
selection.on("mousedown.drag", listener); // re-added
```

burner986:
@mbostock Sorry i won't do that again. And thanks so much that did the trick.

Issue #1888, opened 2014-05-25
Hierarchy layouts need better tests.


mbostock:
Should at least have flare.json-based tests. Tests for the sticky treemap layout would also be useful and have caught the 3.4.7 bug.

mbostock:
Implemented in [d3-hierarchy](https://github.com/d3/d3-hierarchy/blob/master/test/treemap/flare-test.js).

PR #1891, opened 2014-05-28
re-did hierarchy to have flare.json based tests.

MichaelHilton:
Updated the tests for hierarchy to have them be flare.json based as per #1888.

Not sure if the data should be inline or not.


mbostock:
I’ve implemented tests using the Flare dataset in [d3-hierarchy](https://github.com/d3/d3-hierarchy/blob/master/test/treemap/flare-test.js) for 4.0.

Issue #866, opened 2012-10-22
Transitions do not support "get" mode for attr, style, etc.


ZJONSSON:
I'm not sure whether this is a bug or a feature, but I see the following issue:

```js
d3.select("#rect").attr("height") 
```
returns the height of the selected object

```js
d3.select("#rect").transition().duration(1000).attr("height")
```

does not. I get "Uncaught Error: INVALID_CHARACTER_ERR: DOM Exception 5 "  d3.js line 4405 (v.3)

I understand that this type of usage might be rare, but it might still be relevant:  Custom functions that accept single select() object as an input where (i) attributes of each will both be read and set (with the embedded transition in place if applicable) and (ii) it's not clear from the outset exactly which attributes will be set (hence the custom function).

cpbotha:
I'm inside of a d3.svg.axis() automatic transition callback and I'm having trouble getting the final value of the tick transform. (I need to add more labels that are based on the final positions of the automatically animated axis tick transitions)

Based on this issue, als PR1912 and the d3 transition docs, I have not yet been able to figure out how to get these ending / target transform values. Could anyone point me in the right direction?

I have something like:

``` javascript
function myCallback(gTransitions) {
    gTransition.each(function() {
            var g = d3.select(this),
                  ticks = g.selectAll('.tick');

            // ??? here I need to get target / ending value of animated .tick transform
    });

}

var t = svg.transition().duration(750),
      yAxis = d3.svg.axis().and().more();
// later:
t.select('.y.axis').call(yAxis).call(myCallback);
```

mbostock:
I’ve tried to formalize this in [d3-transition](https://github.com/d3/d3-transition) for 4.0. The target-only methods *transition*.style, *transition*.attr and *transition*.text don’t support inspection; there’s no good way to distinguish between an undefined target value and a null target value (other than, say, undefined and null, but that’s dubious). You can inspect the interpolator using *transition*.styleTween and *transition*.attrTween, and you can inspect the tween function using *transition*.tween. So I think that’s good enough.

PR #1912, opened 2014-06-12
Add getters for transition.{attr,style,text}.

jasondavies:
Proof-of-concept for #866.


mbostock:
Decided to not support this in the 4.0 API because there’s no way to disambiguate between null meaning “transitions to the default value” and “no transition is defined for this attribute”. You can however using *transition*.attrTween and *transition*.tween in getter mode.

PR #1964, opened 2014-08-06
Array min&max

Anmo:
Change array.min and array.max to return the element of array when is passed a function.


mbostock:
Moved to d3/d3-arrays#11 for 4.0.

PR #1991, opened 2014-08-26
Quadtree, delete node

ufenegga:
I needed to build a quadtree that can allows to delete/add nodes.

It is a first naive implementation :
- each nodes keep a reference to its parent
- delete method is called on root and accepts only leaf node at the moment
- users should keep a reference to all nodes (typically via the visit method)


mbostock:
The new [d3-quadtree module](https://github.com/d3/d3-quadtree) supports [*quadtree*.remove](https://github.com/d3/d3-quadtree#quadtree_remove) and will be part of [D3 4.0](https://github.com/mbostock/d3/tree/4).

PR #2022, opened 2014-09-17
Zoom behavior with custom scaling

jchlapinski:
I have added support for custom scaling while zooming with `d3.behavior.zoom`. Example and changes in d3 API are described at http://bl.ocks.org/jchlapinski/d51d9aca2084bc02f30b

A method `scaleGenerator` was added to the 'zoom' object as follows: 
```javascript
var zoom = d3.behavior.zoom();

zoom.scaleGenerator(function (scale, delta, usecase) { // below is the default implementation
  if (usecase === "mousewheel") {
      // delta = event.deltaY (or equivalent for different browsers);
      return Math.pow(2, delta * .002) * k;
    } else if (usecase === "dbltap") {
      // delta = 0 
      return k * 2;
    } else if (usecase === "dblclick") { 
      // delta = 1 or -1 if user double-clicks while holding SHIFT
      return Math.pow(2, delta < 0 ? Math.ceil(Math.log(k) / Math.LN2) - 1 : Math.floor(Math.log(k) / Math.LN2) + 1);
    } else if (usecase === "touchzoom") {
      // delta is equal to the distance change between touches (two-finger) in pixels
      return k * delta;
    }
    return k;
});

var sg = zoom.scaleGenerator(); // get current scale generator

zoom.scaleGenerator(null); // restore default scale generator
```


mbostock:
As with #2605 this feels to me like something that should probably be fixed internally, rather than exposed as a new public API.

Related #2426 #2589.

mbostock:
This is a duplicate of #2605, and probably not related to just fixing mousewheel events on Windows, since it’s more about adding greater control over the zoom behavior.

My first concern with both of these proposals is whether it’s better to intercept the input events yourself and then change the scale of the zoom behavior, rather than letting the zoom behavior handle the input events while exposing hooks for how the zoom behavior computes the new zoom level.

For example, maybe it makes sense to change the behavior of dblclick, since the default of doubling or halving is arbitrary. But it’s not hard for you to intercept the dblclick event and trigger a zoom transition instead of letting the zoom behavior handle it.

And, it probably doesn’t make sense to override the interpretation of touch gestures; those are designed such that the touch point coordinates remain consistent under pinch and pan.

Similarly, it’s questionable to change the behavior of mousewheel, since consistency is good for usability (aside from it being broken on Windows in some cases)… but sure, maybe it makes sense to override in some contexts for slower or faster zooming. You could also intercept the mousewheel events and do this outside the zoom behavior, though admittedly this is a bit harder to do than intercepting dblclick because a sequence of mousewheel events is typically aggregated into a zoom gesture (zoomstart, zoom+, zoomend) using a timer.

jchlapinski:
Well, this PR was created because I needed precise control of scale factor for one of my projects. My particular use-case was that I charted some financial data on a bar chart with time scale and then used some non-linear aggregations while zooming out or in; When zooming out the scale was represented in months, quarters, 6 months, years, 5 years and decades. I wanted to precisely control consecutive scaling factors especially for mouse-wheel zooming to achieve some nice animations - three month bars would gradually become one quarter bar, etc. The point of adding proposed API was that it currently is not possible to scale for exactly 1/3 after single mouse wheel event, and then scale for 1/2 after another.

I did not test nor think about using touch gestures. I agree that with pinching this approach might need some more thought.

IMO zoom behavior should never expose too much of its internals (like event handling), but it could allow for applying custom scaling factors. 

I do feel however, that my proposed API is clumsy and non-elegant as proposed above and I am open to suggestions how this could look like. Perhaps this should be divided into several scaling factor generator functions for different zooming events (mouse wheel zooming vs pinch zooming vs double tap zooming)? That way this clumsy 'usecase' parameter could be removed from the API?

mbostock:
That sounds like a reasonable approach for discrete input events like dblclick, but I’m not sure it’s appropriate for mousewheel, which is a continuous gesture on some platforms (and some input devices), such as OS X with a Magic Mouse or a MacBook trackpad.

You can, of course, group a sequence of mousewheel events into a single discrete gesture, and this is exactly what Google Maps did before the great switch from raster to vector tiles. But we’d need to change how the zoom behavior interprets the sequence of mousewheel events if you want it to snap to specific zoom levels.

It sounds like maybe the API you want is the ability to specify specific allowed zoom levels, rather than allowing continuous zooming between any zoom level (within the scaleExtent).

Also, I wonder whether you’d want to temporarily allow being between the specified zoom levels, so as to make the transition between allowed zoom levels smooth and continuous.

jchlapinski:
I also thought at first about just supplying a discreet array of scaling factors for scaleExtent, however as much as this would more or less suit my particular use-case at the time I thought it is much better to be able to supply a function, since that way one can do some dynamic and continuous calculations dependent on a specific context. Just much more general and also very easy to support discreet values case.

Also I indeed used a continuous scaling so that when zoomed out the chart was scaled smoothly to the next aggregation level. It was a matter of controlling what those end-point scaling factors might be depending on the current aggregation level as they were non-linear.
 
As for @mbostock comment above, I do not understand why supplying custom scaling function would not be suitable for continuous mousewheel  gestures? With appropriate context supplied to the scaling function it could always at least behave exactly as the algorithm that is currently implemented, no?.

It seems to me that this scaling factor control is sometimes useful for some users (judging from those few PRs), although probably not that often, and if done well it should be absolutely harmless and unobtrusive to all. 
Is it a worthy feature to add to API? I do not know, it is definitely nonessential, but I remember I felt a little bit disappointed that I could not do what I wanted with zoom behavior. You can control almost every computational aspect through this beautiful and simplistic API in d3, why can't You control just that?

jchlapinski:
> My first concern with both of these proposals is whether it’s better to intercept the input events yourself and then change the scale of the zoom behavior, rather than letting the zoom behavior handle the input events while exposing hooks for how the zoom behavior computes the new zoom level.

Well when I think of it now, it does seem to be relatively easy to intercept input events, if slightly more difficult with the mousewheel, and simply use zoom.scale() to achieve custom scaling... 
There is quite some code in zoom behavior source that deals exclusively with input events, so re-implementing that is not trivial but one can always reach for the code provided there to speed things up. It just seemed much easier to me at the time to patch d3 instead of re-implementing it altogether but now I am not sure whether this is worth it and whether there is a really good way of customizing scaling algorithms in that manner... After all if one wanted to do some more animation oriented and dynamic zooming it would probably be impossible to do properly without intercepting input events anyway.

If Your @mbostock concern is that this feature would bloat the API, which is a very important concern IMO, I will not insist on including this feature. I will close this PR for now, since the code needed work anyway. Thanks for the discussion and for Your insights.

PR #2090, opened 2014-10-26
Optimize d3.keys

jonsadka:
Native JavaScript Object.keys() method is heavily optimized for the specified task (http://jsperf.com/object-keys-iteration/20).  The native method does not perform on the object's prototype and as such, I have concatenated the object's keys with that of the prototype's keys because based on previous implementation it seems the keys of the prototype is also desired. Concatenation is also more optimized than pushing (http://jsperf.com/concatperftest/6).


mbostock:
Sorry it’s taken me so long to look at this, and thanks for the suggested performance improvement.

There’s a subtle issue here in that you can’t just call Object.getPrototypeOf once—you have to crawl up the entire prototype chain to pick up all the inherited properties that would be enumerated by for…in. That still [appears 30% faster](https://jsperf.com/for-in-vs-keys/3), at least for this one test case (that may not be representative of real-world usage).

But, the bigger issue for me is that the code is certainly more complicated. I can’t see this code being a performance bottleneck, so I’d favor the simpler code.

PR #2091, opened 2014-10-26
layout/force: handle case of zero weight on both linked nodes

alon:
In this case treat both nodes as having an equal weight of 1
(arbitrary).

This fixes a divide by zero error otherwise.

I've included a signed-off (saying that I'm the author of the code). I didn't sign the contributor agreement since this falls under the trivial rule. Correct me if I am mistaken.


mbostock:
This fix doesn’t handle the case where only one of *s* or *t* has a zero weight (it treats the zero weight as one), but that’s easy enough to fix. :grin:

mbostock:
Merged into #2591 for 3.5.7.

PR #2102, opened 2014-11-04
Use .attr() instead of .style() for axis label's "text-anchor" assignment.

floledermann:
This allows the label's text-anchor setting to be overridden by CSS without having to resort to `!important`-statements (which are ugly and a maintenance hell) - for example to create horizontally centered labels on vertical axes. You are using `attr()` for visual attributes pretty much everywhere else in the code, so maybe there's some reason I've overlooked for specifically using `style()` here, but from what I see there should be no reason to not use `attr()` for assigning the `text-anchor` value.


floledermann:
The patch doesn't merge without conflicts any more, but should be trivial to adapt. Let me know if you need an updated version that merges with current master.
Btw. I signed the contributor agreement when I first submitted the pull request.

zackd:
+1

Wooyol:
NAVER - http://www.naver.com/
--------------------------------------------

neolaocoon@naver.com 님께 보내신 메일 <Re: [d3] Use .attr() instead of .style() for axis label's "text-anchor" assignment. (#2102)> 이 다음과 같은 이유로 전송 실패했습니다.

--------------------------------------------

받는 사람이 회원님의 메일을 수신차단 하였습니다. 


--------------------------------------------

mbostock:
I like this idea. It’d be nice to extend it to include default styles for the ticks and domain path:

```css
.axis line {
  stroke: #000;
  shape-rendering: crispEdges;
}

.axis path {
  fill: none;
  stroke: #000;
  shape-rendering: crispEdges;
}
```

(I’m not 100% certain crispEdges should be included by default, yet.)

It pains me when I see someone forget to style the axis:

screen shot 2016-05-08 at 8 30 17 am

However, these changes are not backwards-compatible, since it changes the default appearance. So this will be part of D3 4.0; see d3/d3-axis#7.

mbostock:
Fixed for D3 4.0.

PR #2121, opened 2014-11-17
add Japanese translation.

visvirial:


mbostock:
I recently merged a slightly different ja-JP locale in 28acd11a242245b8bd32cfcdaaa8d7d382c6272f. If that’s not correct, please let me know and I will make the necessary edits. Thank you!

PR #2125, opened 2014-11-22
Create chinese translation for d3-csv.txt

ganchengyuan1990:
I made a chinese translation for d3-csv(https://github.com/mbostock/d3/wiki/CSV).


mbostock:
Would you mind making this edit directly on the wiki? It is publicly-editable. (I had to lock it temporarily due to spam, but it’s public once again.)

PR #2150, opened 2014-12-10
Fix clearTimeout usage

turbolent:
While working on reimplementing most of the Web APIs required by D3 so it can be used in iOS apps 
(see https://github.com/turbolent/d2a), I noticed that `clearTimeout` is used incorrectly. Its definition is `void clearTimeout(long handle);`, which means that it should only be called with a timer ID obtained from `setTimeout` and not with `undefined`.


mbostock:
This is fixed already in d3-timer (see [timer.js](https://github.com/d3/d3-timer/blob/master/src/timer.js)) for 4.0. It’s probably not worth fixing for 3.x. Even though it’s a strict violation of the API, `setTimeout(undefined)` is a no-op in every supported environment, as far as I am aware.

turbolent:
:+1: Looks good in d3-timer! Looking forward to the modularized 4.0, thanks for the amazing work!

PR #2151, opened 2014-12-10
Implemented ISO 8601 %V directive in time format

eric-poitras:
I have implemented the %V directive as defined in the documentation of strftime. It is following the ISO 8601 standard and it works for formatting and parsing. I also added tests for corner cases like the 2009 year that has 53 weeks.

Parsing note: If you parse something like this : "Fri 53 2009", it yield a date in 2010 since week 53 span over the next yea. Same thing, if you parse "Wed 01 2009", it yield a date in 2008 since week 1 start in previous year.


eric-poitras:
Hi. I was wondering if you plan to give me feedback over this pull request. We use d3 in our product and are very motivated to contribute to the project by fixing issues / adding features directly to the library when appropriate. So I signed the contributor form and I tried to follow as close a possible to the library standards. Is there additional steps I could take to make this merged in the library ?

zefrog:
:+1: This would be a nice feature, as I am using ISO week numbers in tickFormat for some graphs, and starting at zero makes no sense...

eric-poitras:
Merged back master into my branch, and resolved the conflict. Had to fix the bin/component file that was referencing a non-existent attribute. Also, there is a test that fails in mbostock/master so it fail in my branch as well.

mbostock:
Filed d3/d3-time-format#8.

tdc-studio:
I am developing with d3 v4.9.1, can I ask if this very helpful feature of ISO Week numbers is available to me please? If I try and format my ticks using `%V` it doesn't work, unlike `%W` which of course does. Have I completely missed something obvious.

PR #2167, opened 2014-12-22
Use createElement if namespaceURI is undefined.

### LINKED ISSUE #2161: D3 and Shadow DOM
@mbostock @jasondavies 

There seems to be some strange cases where D3 doesn't work properly with the Shadow DOM. I've created an example page to demonstrate one (`.append('textarea')`) of them here: 

http://bl.ocks.org/pemrouz/4a3e7ce9322c091ec058

mbostock:
Normally selection.append is only called on normal elements that have a defined namespaceURI. However, if createElement is called on a shadow root, then namespaceURI is null. Now we correctly fallback to createElement instead of createElementNS in this case. Fixes #2161.


mbostock:
Merged into #2227 for 3.5.4.

Issue #1453, opened 2013-08-12
Winding order determination should be rotation-invariant.


mbostock:
Received an interesting case where the Mercator projection’s default clipExtent exhibited a winding order bug:

[![screen shot 2013-08-12 at 10 57 47 am](https://f.cloud.github.com/assets/230541/949199/cc04470a-0378-11e3-9292-de962d473b0d.png)](http://bl.ocks.org/mbostock/0cc9d3f427dbb9cc86ef)

Simply changing the clipExtent to crop to the display area fixed the problem:

[![screen shot 2013-08-12 at 10 58 02 am](https://f.cloud.github.com/assets/230541/949201/d668b230-0378-11e3-9106-4f8d3c14a742.png)](http://bl.ocks.org/mbostock/a61eb91b975ec1d37aed)

/cc @jasondavies

PR #2169, opened 2014-12-23
Decouple rotation from clipping.

mbostock:
Spherical rotation is now applied prior to spherical clipping, eliminating the requirement for spherical clipping to perform the rotation.


jasondavies:
I’m happy with this *in principle*, but the original reason for the coupling was that rotation of a degenerate polygon might cause it to be treated differently for winding order detection: #1453.  In other words, if we clip unrotated polygons, at least we are consistent with the winding order detection used by TopoJSON.

I’m optimistic that we can eventually rotate any polygons prior to clipping, when I finally finish my improved clipping/point-in-polygon implementation; it might be better to move this to the 4.0 milestone!

mbostock:
Thanks for the context. I had forgotten… Basically this is just a revert of 31ea7dc0f1ff553c446d91591826ec2731fadff1.

I am happy putting this in the icebox. I want to think about some larger things for 4.0 besides.

PR #2170, opened 2014-12-23
Fix quadtree find with accessors

tpreusse:
`d3_geom_quadtreeFind` was using the node data instead of `x` and `y` properties, making it only useable with position arrays.

```
  ✓ OK » 2687 honored (9.716s) 
```


mbostock:
Thank you! Merged into #2227 for 3.5.4.

PR #2216, opened 2015-01-31
Fix multiple drag behaviours on single element

RudolfHattenkofer:
Hey! Not sure if there is a specific reason this didn't work before, but if you make these modifications you can have multiple drag behaviour like this:

    a = d3.behavior.drag('drag-a').on( ... );
    b = d3.behavior.drag('drag-b').on( ... );
    a.call(element);
    b.call(element);

If this is of any use for you I can compile the source files properly!

### DIFF
--- a/src/behavior/drag.js
+++ b/src/behavior/drag.js
@@ -6,7 +6,7 @@ import "../event/mouse";
 import "../event/touch";
 import "behavior";
 
-d3.behavior.drag = function() {
+d3.behavior.drag = function( event_id ) {
   var event = d3_eventDispatch(drag, "drag", "dragstart", "dragend"),
       origin = null,
       mousedown = dragstart(d3_noop, d3.mouse, d3_behavior_dragMouseSubject, "mousemove", "mouseup"),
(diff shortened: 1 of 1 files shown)


mbostock:
Interesting. Can you elaborate a bit on why you need this functionality?

RudolfHattenkofer:
Multiple drag behaviours on the same element won't work, because they all get the same identifier for the mouse events ;) See this example: http://jsfiddle.net/jsnqu5u0/

mbostock:
Yes. But why do you need multiple drag behaviors on the same element?

RudolfHattenkofer:
Well, the short answer is: Because modularity :D But seriously, there is going on a lot in my app (e.g. Panning/Zooming, moving parent elements, moving children, rotating stuff...), I can't just define one single correct behaviour for most elements.

ryank109:
Any update on this?
I would want this for modularity as well.  I have a group of elements that the user can drag and resize.  It would be nice if I could break up the functionality for just moving and resizing.

rhardih:
I've run into another use case, also related to modularity. I've extracted a "selection box" a'la http://bl.ocks.org/lgersman/5311083, into it's own component plugin, with a simple interface:

```js
var svg = d3.select('svg#foo');
var sb = d3.selection_box();
svg.call(sb);
```

Since the selection box internally has to attach a drag handler to show and update it's representant rectangle, if I ever attach a drag handler to my `svg` selection from the outside, the one already attached by the selection box get's overwritten, and as such it stops working.

fent:
Came here for the same use case, modularity. If not able to attach a new drag behavior, where do I get the active behavior?

PR #2218, opened 2015-02-03
Add verbose documentation to the circle pack layout.

ericandrewlewis:
Took me a while to understand how the circle pack layout algorithm works. I modified and added documentation to explain what's happening internally. 

If this is a desirable direction for granular documentation here, it could benefit from another pass by someone with more familiarity.

### DIFF
--- a/src/layout/pack.js
+++ b/src/layout/pack.js
@@ -1,6 +1,11 @@
 import "layout";
 import "hierarchy";
 
+/**
+ * @see <a href="http://portal.acm.org/citation.cfm?id=1124772.1124851"
+ * >"Visualization of large hierarchical data by circle packing"</a> by W. Wang,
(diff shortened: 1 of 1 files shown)


mbostock:
Thanks for this suggestion. However, I’ve subsequently re-implemented the pack layout for D3 4.0 in [d3-hierarchy](https://github.com/d3/d3-hierarchy). There were some bugs due to a misunderstanding (and / or poor wording) of the [Wang *et al.* paper](https://dl.acm.org/citation.cfm?id=1124851)—which is also paywalled. You can find the new circle-packing code here:

https://github.com/d3/d3-hierarchy/blob/master/src/pack/siblings.js

There are a few comments, but I tend to be fairly frugal with comments, favoring essays, examples, and documentation over in-code comments.

I’m planning on writing up a better visual explanation of the algorithm, but I need to work on getting D3 4.0 released first. Thanks!

PR #2222, opened 2015-02-06
Make named colors case-insensitive

wilg:
[Looks like HTML color names are case-insensitive.](http://www.w3.org/TR/html4/types.html#h-6.5) This patch ensures that they're interpreted correctly regardless of case.


src/color/rgb.js
```diff
@@ -86,7 +86,9 @@ function d3_rgb_parse(format, rgb, hsl) {
   }
 
   /* Named colors. */
-  if (color = d3_rgb_names.get(format)) return rgb(color.r, color.g, color.b);
+  if (typeof format === "string") {
```
mbostock: The type of `format` is guaranteed to be a string here, so this check is redundant.

mbostock:
Merged into #2227 for 3.5.4.

PR #2223, opened 2015-02-06
Fix #1491: d3.transition() broken in IE9.

### LINKED ISSUE #1491: Object doesn't support property or method 'transition'  in 3.3
World Tour Map not working : http://bl.ocks.org/mbostock/4183330

It's working in IE on version 3.2.8. But not in 3.3. Any suggestion to get this working?

SCRIPT438: Object doesn't support property or method 'transition' 
 d3.v3.min.js, line 4 character 29727

jasondavies:
Due to the lack of support for prototype injection in IE9, we use direct
extension as a fallback.  However, when d3_selectionRoot is declared,
some methods have not yet been added to d3_selectionPrototype, hence
they are not copied to d3_selectionRoot.

Since d3_selectionRoot is only used by d3.selection and d3.transition,
we replace its eager initialisation by a lazy initialisation within
d3.selection, and modify d3.transition to call d3.selection on order to
retrieve the singleton selection object.


andresgutgon:
:+1: Thank you!

mbostock:
Coincidentally I made a similar change in #2225. I think it’s actually confusing that d3.selection() returns the same instance every time, especially given that d3.transition() returns a new instance for each invocation. So I changed it to always call d3.select(document.documentElement).

Issue #2180, opened 2015-01-07
Update to JSDom 1.5.0


peller:
Significant performance improvements since 1.0.0 should help D3, such as NodeList iteration and insertion.  See https://github.com/tmpvar/jsdom/issues/694

mbostock:
Sounds good. Though, I have also been thinking about removing the JSDOM dependency entirely, and instead changing D3 to rely an externally-provided document. This way, you can use D3 with whatever DOM you desire, and Windows users won’t run into so many issues trying to install Contextify.

fczuardi:
not only windows users, any iojs or node unstable users as well, see https://github.com/brianmcd/contextify/issues/149

greim:
D3 is a fairly popular module that won't install on io.js (for reasons mentioned above). Here's a link to io.js's own thread documenting such modules. https://github.com/iojs/io.js/issues/456

mbostock:
Closing in favor of #2225, which demotes JSDOM to a development dependency used only for tests. (I also upgraded to JSDOM 3.x, but it’ll only benefit the tests.)

mbostock:
Fixed in 3.5.4.

Issue #1986, opened 2014-08-21
Make jsdom an optional dependency in package.json


parkermalenke:
*First just wanted to say thanks to everyone involved for working on D3, this is a very cool project and I've enjoyed using it.*

To the point: jsdom is not actually a requirement if you use the npm version of D3 in the browser with a tool like browserify. There's a couple problems with marking it as a dependency in package.json:

- jsdom itself requires contextify, which can be pretty difficult to get set up on the windows platform. Marking the package as optional would let people completely avoid this irrelevant challenge if they just use D3 in the browser.
- Even if contextify installs easily on your platform that's an unnecessary step that introduces extraneous code for projects that only use D3 in the browser.

mbostock:
Fixed in #2225, which demotes JSDOM to a development dependency only used for D3’s tests.

natevw:
Whoohoo, thanks! The way things are now handled (checking nodes to determine their context) sounds excellent.

vinimxit:
What are the commands to install d3 and c3 without dependency on python and contextify?

mbostock:
I can’t comment on c3, but for d3, a simple `npm install d3` should install without any dependencies. Like so:

```bash
$ npm install d3
d3@3.5.5 node_modules/d3
$
```

PR #2225, opened 2015-02-07
Demote JSDOM to development dependency.

### LINKED ISSUE #2190: Remove JSDOM dependency.
Installing JSDOM is a pain on Windows, and there’s no good reason for the Node version of D3 to dependent on this specific DOM implementation. Instead, users should be able to provide their own DOM implementation as desired.

mbostock:
Fixes #2190 #1491 #2180 #1986 #1250.

Code that previously assumed a global document or window now uses the related node’s ownerDocument or ownerDocument.defaultView as appropriate.

If no related node is available, the corresponding code will crash; however, the rest of D3 will work just fine. For example, you can’t use d3.select(string) unless a global document is available; it just doesn’t make sense. Use d3.select(node) instead, followed by selection.select(string).

Code that previously checked for a global on the window (e.g., XDomainRequest) now uses the global context (`this`) rather than the window.


package.json
```diff
@@ -50,14 +50,12 @@
...
-  },
   "devDependencies": {
-    "smash": "~0.0.12",
-    "uglify-js": "2.4.0",
-    "vows": "0.7.0",
-    "seedrandom": "2.3.1"
+    "jsdom": "3",
+    "seedrandom": "2",
+    "smash": "0.0",
+    "uglify-js": "2.4",
```
jasondavies: Would it be better to pin to a specific UglifyJS patch version, since changes in the patch version can cause d3.min.js to change?  For example, I had to update to 2.4.16 to get d3.min.js to match the build in this pull request.  On the other hand, not pinning to a patch version means less maintenance. :)

src/selection/style.js
```diff
@@ -16,7 +16,10 @@ d3_selectionPrototype.style = function(name, value, priority) {
     }
 
     // For style(string), return the computed style value for the first node.
-    if (n < 2) return d3_window.getComputedStyle(this.node(), null).getPropertyValue(name);
+    if (n < 2) {
+      var node = this.node();
+      return node.ownerDocument.defaultView.getComputedStyle(node, null).getPropertyValue(name);
```
jasondavies: I appreciate this might be for performance, but should `d3_window(node)` be used here instead of `node.ownerDocument.defaultView`?

jasondavies:
Looks great!

jasondavies:
Fix added for `d3.select(window)` bug.

package.json
mbostock: Sure, I’ll restore a specific version for UglifyJS. Unpinning doesn’t really buy us anything since you still have to remember to `npm update` to grab the newer version.

src/selection/style.js
mbostock: I had this exact change in my working tree but I forgot to commit it last night, so thanks!

mbostock:
Fixed in #2227 for 3.5.4.

klokoy:
Should´t this be considered a braking change? When using d3 in node the following used to work

```
var d3 = require('d3');

var svg = d3.select('body')
            .append('svg');
```
As jsdom is not included in d3 this will no longer work.  It was a very nice way to generate svg on the server.

mbostock:
It wasn’t explicitly specified that this was supported, and given the difficulties users were seeing installing D3 on Windows (and io.js), I wanted to make the change now rather than wait until the next major release.

You can fix your code as follows:

```js
var d3 = require("d3"),
    jsdom = require("jsdom");

var document = jsdom.jsdom("<html><head></head><body></body></html>"),
    svg = d3.select(document.body).append("svg");
```

klokoy:
Fair enough.  And thanks for your fix.

Kim

toolness:
Thanks for this!

I noticed that the D3 wiki docs on [Browser / Platform Support](https://github.com/mbostock/d3/wiki#browser--platform-support) still said JSDOM was included with D3, so I [updated them](https://github.com/mbostock/d3/wiki/Home/_compare/3bdf0b056e7f0f4e089e02b15193f5ef2a52b0eb...900dc4a298abf58f900ba4da5c4b5e993ffaf6b1) to reflect this change... If you don't like it though, feel free to change it.

mbostock:
Thank you!

PR #2228, opened 2015-02-08
Simpler no-global for CommonJS.

mbostock:
We don’t need the index.js to unset the global if we don’t set it.


mbostock:
Merged into #2591.

PR #2237, opened 2015-02-16
Extend svg.arc.padAngle parameters.

Caster:
##### Description of the change
The `padAngle` function accepted a single number previously, but now accepts an array of two values too, in which case the left and right pad angle are set separately as specified in the array. This allows for the creation of sunburst charts with straight separation edges. Of course, the `padAngle` can still return an accessor function, that can return an array.

##### Reason for the change
This was something I needed in a project, but it was not possible to do this without changing D3 code. Hopefully, my changes can (after some feedback) be of use to others too.

The fundamental problem is that when one applies a gap between arcs in a partition using some definition in radians, a wedge-shaped gap is formed. The existing `padAngle` implementation handles this nicely already, by making the inner and outer padding slightly different, so that a straight edge gap is formed. However, it is not possible to extend this gap to outer rings, because a segment always has a padding both to the left and right of it. This change makes it possible to set the padding for the left and right side of a segment separately.

##### Demo / application of the change
[This bl.ocks.org demo](http://bl.ocks.org/Caster/e5b50054e36a34dbb6b8) demonstrates what I needed in my project, using some made up data. Notice how the padding between segments in the inner ring extends to the outer ring, while the segments in the outer ring do not have inner padding.

##### Remarks
I noticed how there is *a lot* of functionality stuffed in that function already ( :-) ), so some things may not work well together with this change (like rounded segment corners). However, all existing tests still pass. I am just not really sure how to write tests, so I would really appreciate some pointers if tests are desired (and I assume they are).

All feedback is appreciated and I am definitely willing to make (big) changes if necessary :-)


mbostock:
I took a stab at this for 4.0 in d3/d3-shape#42. Before:

screen shot 2015-12-01 at 10 48 36 am

After:

screen shot 2015-12-01 at 11 06 24 am

There’s still some work to-do, and I need to think through whether this is the right API. Also, there’s the issue that applying constant padding to arcs disproportionately penalizes small arcs, so it makes me wonder if the partition layout should support padding so as to preserve relative areas. (See d3/d3-hierarchy#19 for a description of a similar problem with the treemap layout.)

Caster:
Thank you for considering this change!

Indeed the API could be better, as it was for personal use only I just went with this. Very interesting to see the changes you are making, it looks promising. I agree with the penalizing of small arcs, it would be awesome to have D3 take care of that. At the moment I am kind of busy, but if there is anything I can help with, tinker around a bit, just let me know :-)

mbostock:
Merging into d3/d3-shape#42.

PR #2238, opened 2015-02-16
Skip creating half edges when we don't have a second vertex

portante:
Running into problems with voronoi use from with the nvd3 library, and while debugging I noticed that cascading conditional starting on line 51 of `cell.js` can end up with a `null` value, which causes `CreateBorderEdge` to choke de-referencing for an `x` value.

Not really understanding this code, it seems that we would not want to create a border edge in that case.

Is there any harm it just skipping its creation?

FWIW, the nvd3 code was fixed to remove duplicate points (see https://github.com/novus/nvd3/pull/584).

Thanks for any insights you can offer.


xiashuyan:
Rrr

portante:
@mbostock, could you take a moment to consider this?  Thanks.

mbostock:
Looks like it chokes in `d3_geom_voronoiHalfEdge`, but yeah. Needs further investigation.

PR #2266, opened 2015-03-20
Update attr.js

johnstark:
Prevents error that is thrown when a resize event happens.


johnstark:
Prevents error that is thrown when a resize event happens.

jasondavies:
Sorry, there’s not enough information here to understand what error this is fixing.  It sounds like the issue should be fixed outside of D3, i.e. you shouldn’t be setting an attribute to NaN on resize.

PR #2351, opened 2015-03-24
Fix combined mousewheel/drag gesture.

### LINKED ISSUE #2205: Problem with d3.behavior.zoom (Mac OS)
The page loads, then you pan to either side and then zoom, and it throws a javascript error in console. This is strange because just today I was using zoom behavior in Windows, Chrome and (as far as I remember) it worked.

__The problem only exists if you use 3-finger trackpad gesture for dragging and don't change cursor position after panning.__ Doesn't work if you drag with hard-click. So, pan with 3 fingers and try to zoom right away. Seems like d3 can't determine cursor position after dragging.

Errors
Safari: `TypeError: null is not an object (evaluating 'p[0]')` line 1228
Chrome:
```
Uncaught TypeError: Cannot read property '0' of null    d3.js:1382

translateTo    d3.js:1382
mousewheeled    d3.js:1519
(anonymous function)    d3.js:1084
```
Tested in both browsers on Mac OS.

Addition:
Recreated the same in Firefox. Firefox and Chrome are pointing to this bit of code (lines 1380-1384):
```js
function translateTo(p, l) {
      l = point(l);
      view.x += p[0] - l[0];
      view.y += p[1] - l[1];
    }
```
Here's a short example:

```html
<!doctype html>
<html>
    <head>
        <script type="text/javascript" src="d3.js"></script>
        <script type="text/javascript">
            window.onload = function () {
                var svg = d3.select("body")
                    .append("svg")
                    .attr({
                        width: 640,
                        height: 480
                    });
                var data = d3.range(10);
                var x = d3.scale.linear()
                    .domain(d3.extent(data))
                    .range([30, 610]);
                var zoom = d3.behavior.zoom()
                    .x(x)
                    .scaleExtent([1, 10])
                    .on("zoom", function () {
                        d3.selectAll("circle").attr("cx", function (d) { return x(d); });
                    });
                var g = svg.append("g").call(zoom);
                g.append("rect")
                    .attr({
                        x: 0, y: 0,
                        width: 640,
                        height: 480,
                        opacity: 0
                    })
                g.selectAll("circle")
                    .data(data).enter()
                    .append("circle")
                    .attr({
                        cx: function (d) { return x(d); },
                        cy: 50,
                        r: 10,
                        stroke: "none",
                        fill: "black"
                    });
            }
        </script>
    </head>
    <body></body>
</html>
```

jasondavies:
During a combined zoom/pan gesture, controlled via simultaneous use of
the mousewheel and dragging, it is possible for the drag to finish
before the use of the mousewheel finishes.  The mousewheel gesture is
deemed to be finished if 50ms passes with no mousewheel event.

Although the "zoomend" event is only dispatched when *both* gestures
have finished, the variable "center0" was being reset when *either*
gesture was finished, causing an error in the mousewheel handler.

Fixes #2205.


aubergene:
I'm still getting an issue when using this fix. With this [example](http://bl.ocks.org/mbostock/9656675) I get this error. Using Chrome 41 on OSX with an Apple Magic Mouse. I click a state and then scroll slightly whilst it is transitioning, which is very easy to do, since the mouse is very sensitive.

```
Uncaught TypeError: Cannot read property '0' of nulltranslateTo @ d3.js:1427
mousewheeled @ d3.js:1564
(anonymous function) @ d3.js:1120
```

jasondavies:
Are you sure you’re testing the patch?  It hasn’t made it to a release yet.  The call to translateTo is on line 1564 in master, but 1563 in the fix-zoom branch.  Your error message seems to indicate you’re testing master.

aubergene:
I manually applied the patch to d3 3.5.5. I've made a [gist here with the issue using the patch](http://bl.ocks.org/aubergene/8be421c35ed71494d27a).

jasondavies:
Great, thanks.  I’ve pushed a fix for that issue too now.  The issue was that interrupting an in-progress transition caused `center0` to be reset, so it needed to be assigned after the interruption.

PR #2356, opened 2015-03-26
Remove attribute when the value x is NaN

hyunkim9123:
I found a un-caught exception regarding to NaN while following tutorial part 3.
The root cause of this is a parse error, and this occurs when the data file is not properly formatted.
When attempting to set attribute, we already avoid when the value x is NULL.
By removing attribute when the x is NaN, we can prevent un-handled NaN exception in browser.

### DIFF
--- a/src/selection/attr.js
+++ b/src/selection/attr.js
@@ -46,7 +46,7 @@ function d3_selection_attr(name, value) {
   // or remove the attribute as appropriate.
   function attrFunction() {
     var x = value.apply(this, arguments);
-    if (x == null) this.removeAttribute(name);
+    if (x == null || isNaN(x)) this.removeAttribute(name);
     else this.setAttribute(name, x);
   }
   function attrFunctionNS() {


jasondavies:
I can’t see any browser exceptions when using element.setAttribute("foo", NaN).  Which browser are you using?

hyunkim9123:
I have checked the exception in developer plugin in both chrome and firefox browser. 
If you would like to see the exception, please set the data.tsv below and follow the instructions at http://bl.ocks.org/mbostock/3885304. "data.tsv" is expected to have "tab" as the separator, but I have used "space" instead of that, which causes parsing error and NaN exception.
letter frequency
A .08167
B .01492

jasondavies:
OK, I see.  This is not an exception per se, but more of a warning, which is logged to the console if an invalid SVG attribute is set.  In fact, I think it’s actually a good thing, because it’s easier to figure out why things are not working.  In other words, we know that the chart is not displaying properly, because some values are NaN, which probably means a parsing issue.

The fact that `null` is treated specially (it means removal of attributes) doesn’t necessarily mean we should extend this special treatment to other values, such as `NaN`.  In fact, `.attr(<name>, null)` is part of the API, and I don’t think it would be sensible to make the API more complicated by making other values do the same thing.

PR #2357, opened 2015-03-26
Be nicer to minifiers that do not handle + ++ correctly

tobgu:
Also improves the readability a bit I think.


jasondavies:
Duplicate of #1916.  The file `d3.js` is actually automatically generated by UglifyJS, so the proposed patch wouldn’t help, in any case.  But I think it would be better to fix the minifier, rather than include workarounds.

PR #2360, opened 2015-03-26
Ignore case when detecting colors for interpolation.

### LINKED ISSUE #2359: Color Transition (Chrome Version 41.0.2272.101 (64-bit)) Issue
I noticed that transition for a background-color attribute does not work with some colors.

``` javascript

d3.select('svg')
	.transition()
	.duration(5000)
	.style('background-color', 'SteelBlue');
```

The color appears but does not transition
If you replace the color string to say yellow it does work as expected.

Kinda silly but still wanted to know if there was a reason this (bug) is occurring.
Has anyone ever seen this?

jasondavies:
Fixes #2359.


mbostock:
Do we need a `/i` flag on the rgb/hsl regular expression, too?

jasondavies:
Yes, good point.  I’ve updated the patch to support case-insensitive RGB/HSL functional notation as defined in the CSS spec.

PR #2376, opened 2015-03-31
Fix `d3.selectAll(HTMLCollection)`

natevw:
Using a selection created via e.g. `var sel = d3.selectAll(someElement.children)` used to work, but in https://github.com/mbostock/d3/commit/9c3df31216aa28d8252b216b558cbb5907480b5b#diff-d93c7c1ccac09d19287c11d2d115758aL61 a critical conversion to `d3_array` was factored out of that path. This resulted in such selections exploding when certain array-dependent methods were used, such as `sel.sort()`.

Simple fix, plus now a test for this code path. (The existing test had only been exercising `sel.selectAll` but not `d3.selectAll`!)


mbostock:
Merged into #2591.

Issue #2189, opened 2015-01-13
Script70 "Access Denied" or "Permission Denied" when animating embed'ed svg elements


dwt:
* The bug manifests itself as 'Script70' Error (error message "Access Denied" or "Permission Denied")
* If an SVG element is ```<embedd src=some.svg>``` on a page and is then removed while an animation is still running in it - an exception is sometimes thrown from ```d3.timer.tick()``` or ```d3.timer.stop()```.  To be more specific d3.timer stores state on a DOM node in the __transition__ attribute and then tries to ```delete node.__transition__``` it in ```d3.timer.stop()```. That will fail if the SVG element cannot be accessed (attribute access or assignments fail with an error). If ```d3.timer.tick()``` calls a function that does the animation (this could also be a custom attrTween) and it throws the timer continuation is stopped.
* This is non recoverable by the application and will stop all concurrently runnning animations and will prevent all animations scheduled after that point!
* This is also not reliably preventable by the application as there is no way to synchronously stop an animation. ```selection.interrupt().transition()``` and ```d3.timer.flush()``` all are asynchronous and thus the race condition between stopping the animation and IE killing of the SVG element remains.
* We suspect that the IE js garbage collector is to blame for this, as the problem happens after varying intervals and can be triggered more easily if there is more activity in the DOM.

Possible solutions / workarounds:
* Keeping a reference to the embed DOM element seems to keep it alive - keeping a d3.selection of the SVG element (path, group, etc) inside it doesn't. This could mean that changing the way animations remember the elements they're working on could prevent this error.
* Catching and possibly discariding exceptions thrown from the handlers called by d3.timer.tick() and d3.timer.stop() so one erroneous SVG element doesn't kill all other d3 animations.
* Listen to DOM-removal events and internally and synchronously kill timers on the removed elements.
* Providing a solution to synchronously stop an animation. This could allow an app using d3 to work around this problem by stopping the animation before it removes the containing DOM element.

We would prefer a solution that doesn't require the d3 using app to include ie10/11 workarounds.

That being said, we think it would be very beneficial if we could force d3 to synchronously get rid of timeouts/ transitions/ animations/ ... on dom nodes. This would have helped us tremendously to track down/ work around this problem and would help d3 using applications to more easily work around similar issues when they are found in the future, until a real solution / fix is implemented in d3.

See [this blocks gist](http://bl.ocks.org/rbu/6cda4f71909d0f5e4f11) for a demonstration/ reduction of this problem.

[...]

PR #2377, opened 2015-04-02
transition.remove() fix

robinhouston:
We at Kiln recently noticed that calling `.remove()` on a transition would not always work as expected (though explicitly removing the element from an `.each("end", …)` handler did work).

It turns out that it wasn’t working in the situation where the running transition had preempted a transition that had not yet started (due to a delay).

This patch (with test) seems to resolve the problem, though you may have a better idea for how to deal with it. In any case it seems to me that the `this[ns].count < 2` check is too simplistic, since there may be transitions with ids less than the active one, that will therefore never run.


mbostock:
Interesting. That sounds correct. I should think a bit about whether this is an alternative solution but your approach seems reasonable.

mbostock:
This is related to d3/d3-timer#5 (and d3/d3-transition#9, #2189), in that if we greedily canceled transitions—rather than waiting for them to self-terminate on start—then the count check would be safe.

mbostock:
Fixed in #2600, merged into #2591 for 3.5.7.

Issue #2247, opened 2015-02-27
packed circle charts should have variable padding width, and this should include parent circles


jamiewaese:
I am trying to display a phylogenetic tree with the packed circle chart. It displays the hierarchy of the data structure nicely, and it automatically sets the size of the leaf nodes according to the phylogenetic branch lengths as one would expect.

However, I am not able to adjust the sizes of the parent nodes. This is a shame because they have phylogenetic branch lengths too, just like the leaf nodes. We're masking data here! After a day of hacking, I have to come to realize that there is no clever work-around. 

The pack function allows you to adjust the radius of leaf nodes, but parent node sizes are not adjustable. Also, we can add padding to the leaf nodes but it does not apply to parent nodes.

Ideally we should be able to set the padding for all the circles individually. Parent circles should have an internal padding setting so their children don't automatically touch their sides. They should expand to contain their children *plus* leave a padding gap to display whatever size values are associated with their node.

Any help would be most appreciated!

mbostock:
Moving to d3/d3-hierarchy#7. I also filed d3/d3-hierarchy#9, but the padding solution makes more sense to me than internal node values, since the pack layout won’t be very effective at showing that internal node value (due to “wasted space” between sibling nodes and their enclosing circle).

Issue #1686, opened 2014-01-07
Pack layout padding in parent/child context 


VividD:
If a non-zero padding is applied to pack layout, this will affect minimal distance between siblings. However, a parent circle and its children circles could still touch, or have distance less than padding.

Example: http://jsfiddle.net/VividD/rfBRz/

There are some real world applications that would benefit if padding affected also parent/children distance (so that even a parent and child could never have distance less than padding) (an example would be labeling parents with text flowing along their circles). In fact, my personal impression is that such modified behavior of padding would be more intuitive for developers and users than present behavior.

Perhaps the most desirable resolution would be a new independent parameter called, let's say, parentPadding.

Please consider enabling pack layout to support such modified behavior of padding.

mbostock:
Merging into d3/d3-hierarchy#7.

PR #2379, opened 2015-04-03
Added ability to specify additional size for hierarchical parents

mgetz:
First attempt to make the parents reflect their own size in addition to the size of their children. This is in response to issue #2247 and #1686. This is the same issue I faced as well. I had the additional problem where I could not drag and drop a circle into a parent with a single child since it was the same size.


hughes:
This is great - hope to see it merged soon.

hughes:
Meanwhile, @mbostock has an [example of creating padding with dummy nodes](http://bl.ocks.org/mbostock/3315318)

mgetz:
Thanks for that example, Matt! I had not found that one.

hughes:
It's not a great workaround honestly... the padding is only on one side and the resulting size can be difficult to predict.

mbostock:
Moving to d3/d3-hierarchy#9. I feel like padding (d3/d3-hierarchy#7) is a better solution than internal node value, given the issues with trying to show internal node value meaningfully.

ikari-pl:
This might seem a stupid question, but how can I currently force the `partition` layout, based on hiearchy as well, to use `size` for the parent nodes, *without* switching to `d3-layout`? The padding should be doable and is probably the only option for a sunburst, right?

PR #2381, opened 2015-04-04
Removed bold for D3

Sourdoughh:
D3 was bold when it shouldn't be.


mbostock:
It’s bold because it’s the first use of the alternative, shorter form of “D3.js”.

PR #2383, opened 2015-04-07
Created Korean locale (ko-KR)

e-:


mbostock:
Replacing backslashing with ₩ (presuming intent) and merging. Thank you!

PR #2389, opened 2015-04-10
Refactor interpolators from svg/line to svg/interpolator

### LINKED ISSUE #2385: Exposing line interpolator functions
I'm currently working on a charting library that would benefit from having direct access to the interpolation functions available for d3.svg.line/area. Would it be reasonable to create a pull request that does the following:

1. Break out the interpolators from svg/line.js and puts them into svg/interpolators.js
2. Refactor the default available interpolator functions from `d3_svg_lineLinear` to `d3.svg.interpolators.linear` and so on.

Thoughts?

rcchen:
Move the interpolators from svg/line.js to svg/interpolator.js. Primary purpose is to expose the interpolator functions directly to users. Will update docs if/when this is merged in.

Closes #2385


mbostock:
I don’t plan on undertaking this change for 3.x, but I’ve filed d3/d3-shape#13 for 4.0.

PR #2391, opened 2015-04-13
For people that are using d3 with web components.

boJackEden:
It is simply an import.html with the following content:

```
<script src="d3.min.js" charset="utf-8"></script>
```

When working with web components, every component that uses d3 will import it through:

```
<link rel="import" href="../d3/import.html">
```

This will prevent importing d3 more than once.


mbostock:
Sorry, but this feels like unnecessary baggage and I do not think it is appropriate to put it in the D3 repository. If people are using web components with D3, they should create their own import.html as needed.

PR #2397, opened 2015-04-18
added proposed tickFormat issue fix

MrCox:
I tried to modify the "text-anchor" style for ticks using the `tickFormat` method on d3 axes. The function was roughly this:
```js
function tickFormat(tick) {
  d3(this).style("text-anchor", "end");
  return tick;
}
```

However, the styling doesn't take effect. Internally, d3 adds default "text-anchor" styling to the ticks _after_ the `tickFormat` function is applied, meaning any "text-anchor" value set in the `tickFormat` function gets overridden.

To fix this, I just moved the part applying `tickFormat` to be after the "text-anchor" assignment. This gives a user the ability to override the default styling.

Not a huge deal, but helpful!

### DIFF
--- a/src/svg/axis.js
+++ b/src/svg/axis.js
@@ -44,7 +44,7 @@ d3.svg.axis = function() {
 
       var lineEnter = tickEnter.select("line"),
           lineUpdate = tickUpdate.select("line"),
-          text = tick.select("text").text(tickFormat),
+          text = tick.select("text"),//.text(tickFormat),
           textEnter = tickEnter.select("text"),
           textUpdate = tickUpdate.select("text"),
           sign = orient === "top" || orient === "left" ? -1 : 1,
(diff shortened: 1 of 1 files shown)


mbostock:
Relying on ordering and side effects in the tick format function seems fragile. The behavior isn’t explicitly documented, but I wouldn’t rely on the tick format function having the text element as the `this` context: it’s better if the format function relies only on the passed-in value and returns a string.

If you want to modify the styling of the generated ticks, the recommended pattern is to post-process the axis after it’s rendered: select the ticks, and then modify them. See the [axis styling example](http://bl.ocks.org/mbostock/3371592).

mbostock:
Also, sorry it took me so long to reply, and thank you for submitting this pull request (even if it wasn’t accepted).

Issue #1439, opened 2013-08-06
The zoom behavior should support pointer events.


Ta4i:
Users of Win 8 on touch devices which use IE 10 are disable to zoom in D3 using gestures. It's because in IE there are different gestures model.

Description of Pointer concept:
http://msdn.microsoft.com/en-us/library/ie/hh673557(v=vs.85).aspx http://github.com

Existing approaches of unifying events:
http://blogs.msdn.com/b/eternalcoding/archive/2013/01/16/hand-js-a-polyfill-for-supporting-pointer-events-on-every-browser.aspx

> Back in September 2012, Microsoft proposed a specification to W3C
> for unifying touch, pen and mouse events called Pointer Events
> and based on the APIs available today in IE10 on Windows 8.
> The W3C has since created a new Working Group which already published
> a last call draft.The Microsoft Open Technologies, Inc. team has also released
> a Pointer Events initial prototype for Webkit on HTML5Labs to further contribute
> to the technical discussions in the W3C Pointer Events Working Group. 

So seems like "Pointer" is upcoming standard (at least it's make sense). So it'll be cool to start support this model. Or just add ability to perform zooming by MSGesture.

natevw:
xref: https://github.com/mbostock/d3/issues/2380

bwestpha:
hey @mbostock :)

Since this is a very old request / topic, are there any new information about it? 
It's because Windows tablets are really popular and quite common in big companys and there is no fix for zooming with an MS tablet ?!
Is there any reason why this was going to the icebox?



### (and also, just a question by the way: is there any option / tutorial / code snippet, which makes the tooltips useable with an MS touch device? )

Herst:
Touch events are in Edge but disabled per default and the setting hidden at **about:flags**. With this activated zooming in works but zooming out does not. All-in-all the situation is inadequate.

For what it's worth, here is a simple demo of how to support the pinch zooming gesture for triggering a d3 zoom (not the browser gesture for doing page zooming): <https://mdn.github.io/dom-examples/pointerevents/Pinch_zoom_gestures.html>
Maybe I find time for hacking support for this into d3 because pinch-zooming is the only thing in my project which does not work on touch-enabled devices with Windows (at least if they have keyboards, according to some voice on the web it makes a difference).

[...]

PR #2403, opened 2015-04-22
Pointer events for the drag behavior.

natevw:
This makes e.g. [the bubble demo](http://bl.ocks.org/mbostock/9631744) usable via touchscreen on Windows machines, and should remain future proof as other major browsers adopt Pointer Events (and as MS [adds Touch Events](http://blogs.msdn.com/b/ie/archive/2014/07/31/the-mobile-web-should-just-work-for-everyone.aspx) for backwards compatibility on some of their platforms).

## Miscellaneous notes

- Unfortunately, this required an (albeit minor and backwards compatible) addition to d3.mouse. My [original go at this](https://github.com/mbostock/d3/commit/736cc499739373d861d625affbf87991a36736f5) compared the ids itself, which seemed a better encapsulation, but the current drag code now already expects a missing (i.e. falsy) position return value as d3.touch gives so I've followed suit here. May actually be useful to other PE-aware users of d3.mouse I suppose.

- I've made the touch-action setting a bit more careful in this go-'round. I again debated whether or not d3 should be setting a style property like this, but a) it's already doing much crazier DOM hacks when needed for other event handling, b) better to have it work by default for the 99% of cases and have anyone who really has some way for touch-action to remain 'auto' and still get useful work out of d3.drag (they can set back themselves).

- Finally, as a more general comment, I haven't updated d3.zoom to work with PE. In my opinion, that behavior should be leveraging the work in d3.drag (+ one further trivial modification to also pass along the dragId with its events) to track touches in a cross-platform way but that needs a bit further discussion beyond https://github.com/mbostock/d3/issues/1793#issuecomment-38082270 — I have long been using `d3.drag` to track touches on a *parent* element (rather than an actually dragged element) and it works fine by simply adjusting via `.origin()`.


lubomirov:
Hi, I have a working version of d3.behavior.zoom that fully support MS pointer events. The patch is local to zoom.js file. It's about +100 source lines. It can be compact, I didn't an optimization so far. The minimized js size is about +2Kb to d3.js. Could we join our efforts in the area of support pointer events in D3?

mbostock:
Fixes #2556.

nikoloas:
Hi I have just gone through the truck load of updates to support touch on win8+IE11 and none of these fixes actually work for me.  All I'd like to have is the pinch zoom and panning that is supported by default on a SVG element and when a zoom behavior is implemented.  I have found that adding CSS touch-event:none on the SVG element I am able to pan, however the pinch zoom does not work.  I am guessing this isn't related to your solution, how would this update affect my existing d3 drawing?  
Many thanks!

natevw:
@nikoloas Sorry I missed your questions earlier. Have you found a solution? If not, be warned this patch is probably way out of date.

The new d3-drag is even more like pointer events, although unfortunately pointer events are still not supported: https://github.com/d3/d3-drag/issues/25

Hopefully this work here could be ported over, I haven't taken a look at what that would involve yet.

Herst:
@nikoloas 

Quoting myself from https://github.com/d3/d3/issues/1439#issuecomment-338496133:
> Touch events are in Edge but disabled per default and the setting hidden at about:flags. With this activated zooming in works but zooming out does not. All-in-all the situation is inadequate which is why there might still be need for the implementation of at least some pointer event stuff.
> 
> For what it's worth, here is a simple demo of how to support the pinch zooming gesture for triggering a programmatic zoom (not the browser gesture for doing page zooming): https://mdn.github.io/dom-examples/pointerevents/Pinch_zoom_gestures.html
Maybe I find time for hacking support for this into d3 because pinch-zooming is the only thing in my project which does not work on touch-enabled devices with Windows.
>
> Currently, I am not very motivated to do so though because in MS Edge 16 (Fall Creators Update from few days ago) the renderer process stops working when zooming or programmatically panning on my page (page randomly freezes and reloads, after some reloads a "This page is having a problem reloading" message pops up. This happens when doing programmatic resizing as well, if anyone else has the problem with a d3-related page please tell me so we can report it to Microsoft).

PR #2405, opened 2015-04-23
Add Pointer Events support to d3.zoom

lubomirov:
Support of pointer events in Internet Explorer 10 and 11


natevw:
Glad to see someone else working on Pointer Event support!

I wonder on this patch, though, if it couldn't be combined more with the existing touch code? This seems to duplicate a lot of the calculations and logic separately for pointer events. I think ideally, a lot of that could be shared with the touch events.

It actually seems to me the zoom behavior might even use `d3.behavior.drag` internally to unify even more handling:

```
d3.behavior.drag().origin(function () {
    // avoid default parent-relative drag positions
    var pt = d3.mouse(this);
    return {x:pt[0], y:pt[1]};
})/* now you have cross-platform pointer tracking via drag{start|move|end} events */;
```

As I mentioned in https://github.com/mbostock/d3/pull/2403 there might need a few little tweaks to the drag behavior to expose a bit more, but then it could abstract out the cross-platform stuff so zoom could do less itself.

lubomirov:
Better version of this pull request was added

PR #2412, opened 2015-05-04
D3 grouping chart

prem-anand:


jasondavies:
We don’t bundle examples with D3; you should put this on <http://bl.ocks.org> instead and perhaps add it to the [gallery](https://github.com/mbostock/d3/wiki/Gallery).

Issue #2413, opened 2015-05-04
Transitions do not pass group index to tween callbacks


natevw:
Transitions usually pass the typical trio of "data, node index, group index" (`d, i, j`) through their operators (`.delay`/`.duration`/`.each`/etc.)

However for tweens the situation is different:

1. For completely custom tweens via `.tween(…)`, the trio is passed to the *factory* callback but only two arguments are passed to the tween function it returns.
2. The `.attrTween` and `.styleTween` pass the first two (data, node index) arguments to the provided tween function followed by an [atypical!] "current attribute value" parameter *instead of* the group index — so their shorthand signature is currently `d, i, a` instead of `d, i, j`.

I've got a patch that fixes the first case, but any simple change to address the missing group indexes in the second case would break backwards compatibility and am not sure the best way to proceed. Perhaps we could check the arity of the tween functions there, passing `j` if `tweenCallback.length >= 4`, but that's an ugly hack at best.

mbostock:
True. I think we can try to make these consistent in 4.0.

natevw:
I'd love to see `attrTween` and `styleTween` made consistent, though it's an unfortunate situation for backwards compatibility given how subtle the break could be — both the old `a` value expected and the new `j` value given could be a reasonable value (e.g. `0`) in casual testing.

Perhaps you could start passing those values with a "deprecation warning" trick in 4.x (as it would break code using `typeof` or whatnot…) and then switch in 5.x or something…

```
function deprecatedNumber(n) {
  var dn = new Number(n);
  dn.valueOf = function () { console.warn("Value will be changing!"); return n; };
  return dn;
}
deprecatedNumber(1) + 1;      // yields `2` after logging warning
```
(Similar for `new String`/`.toString`, and it looks like this would be the only use of `console` so it should probably be gaurded more carefully…or just make any use *throw* instead or warn…)

IPWright83:
There's an interesting relation to https://github.com/mbostock/d3/issues/2246 "Pass DOM element to selection functions" where the group index is mentioned. I've not seen it detailed in an API document anywhere? Am I just missing it as could have a bearing on the implementation of 2246. I guess importantly these 2 requirements should align at the end.

mbostock:
All functions are now consistent in D3 4.0, for both selections and transitions. I’ve standardized on the current datum `d`, the index `i`, and the group nodes array, making the D3 signature similar to array methods such as array.forEach.

PR #2414, opened 2015-05-04
Pass group index ("j") to custom tween functions

natevw:
Note that this is only a partial fix for #2413 — the other part may be "unfixable". However, even if we are stuck with attr/style tweens as they are, I still think it makes sense to fix the custom/generic `.tween` handling for consistency with the rest of D3. I cannot see ever passing a "current value" parameter due to the generic nature of the callback, and IMO best to fill in the "j" parameter now and let any [unlikely?] future parameters fill in *after* that.


mbostock:
All functions are now consistent in D3 4.0, for both selections and transitions. I’ve standardized on the current datum `d`, the index `i`, and the group nodes array, making the D3 signature similar to array methods such as *array*.forEach.

Issue #2415, opened 2015-05-04
Transitions stop working if exception thrown in tween


natevw:
If a tween throws an exception for whatever reason, transitions stop working until the page is refreshed.

The actual problem [besides the code throwing exceptions, of course…] is ± in the [d3.timer](https://github.com/mbostock/d3/blob/9c3df31216aa28d8252b216b558cbb5907480b5b/src/event/timer.js) implementation, however it appears that the transition code is overriding some of that internal mechanism and may also need handling for this.

(I'm generally loath to add too much coddling against "operator error", apart from throwing informative assertions, but in a case of a global handler like this it seems like something that should be fixed. Otherwise any single buggy/careless "user" of `d3.timer` can break it globally for all the others within the page.)

mbostock:
Superseded by d3/d3-transition#48.

mbostock:
(This was sort-of fixed by d3/d3-timer#4 d3/d3-timer#8, in that timers now have well-defined behavior if an error is thrown. But transitions do not yet have well-defined behavior if an error is thrown.)

mbostock:
Related d3/d3-transition@c066c7c7063fcb5a81cdcbfeab5eb6c1502ebbf7.

PR #2416, opened 2015-05-04
Make transitions and d3.timer more robust

natevw:
This addresses some of the places where an exception from a callback can corrupt the internal state of d3.timer and/or d3_transitionNode, resulting in the issue at #2415.

I'm submitting this patch as it does improve the situation, but to be honest it's not a full fix, nor a particularly elegant one. Most of D3 can get away with leaving exceptions to blast their way through, and I think that's a good design. Here, however, seems to be a slippery slope of needing to catch and delayed-throw any possible error to keep the mechanism on track.

The timer/transition logic can get in a really bad state depending on where the exception is thrown. If an exception hits `d3.timer` it simply breaks permanently, no more timers fire whether existing or registered later. Fixing that reveals a worse situation though: the transition logic monkey patches its own (internal to `d3.timer`!) callback, and usually only *after* it has called its own callbacks [tweens, factories, event handlers…] — so the internal transition logic stops advancing properly if *it* doesn't handle exceptions, but it keeps getting triggered triggered triggered triggered forever anyway now that d3.timer does.

Part of me still wants to fix this (thus this start), but I'd like some feedback first on what the best direction is:

- give up and say timer/tween callbacks must not ever throw. *drawback:* if any one caller breaks that rule, all others suffer.
- have d3.timer catch but also automatically cancel any of its callbacks that throw. *drawback:* means transition does not complete, probably leaving its selection/elements corrupted still.
- wrap every single call of d3_transitionNode to any external tween/factory/event(/etc.?) handler with its own try/catch wart
- overhaul d3_transitionNode to be more naturally robust, somehow.

Thoughts?

### DIFF
--- a/src/event/timer.js
+++ b/src/event/timer.js
@@ -52,7 +52,17 @@ function d3_timer_mark() {
   var now = Date.now();
   d3_timer_active = d3_timer_queueHead;
   while (d3_timer_active) {
-    if (now >= d3_timer_active.t) d3_timer_active.f = d3_timer_active.c(now - d3_timer_active.t);
+    if (now >= d3_timer_active.t) try {
+      d3_timer_active.f = d3_timer_active.c(now - d3_timer_active.t);
+    } catch (e) {
+      // we could cancel callback if it throws up, but that ends up breaking
+      // d3_transitionNode's chaining — it is closely coupled to our code.
+      //d3_timer_active.f = true;   // [would] cancel callback
+
+      // we don't want to mask the error, but re-throw it in its own stack.
+      // [`setImmediate` might be better but not worth polyfilling just for this…]
+      setTimeout(function () { throw e; }, 0);
+    }
     d3_timer_active = d3_timer_active.n;
   }
   return now;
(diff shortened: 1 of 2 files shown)


mbostock:
One reason I didn’t do this is fear of performance degradation. I believe try-catch can disable certain optimizations by popular JavaScript engines. And this is the most performance-sensitive part of D3.

Now, I do use try-finally inside event dispatching (see src/event/event.js and src/selection/on.js), so I’m not sure if those optimizations are already precluded, or if it’s better to put the try-catch higher up the stack to allow the inner loop to be optimized. This is all kind of black magic because most of what matters is touching the DOM, unless you have lots of timers.

I haven’t had time to think this through but I also wonder whether it would be possible for the exception to be thrown “naturally”, which could prevent some timers from being called back during that iteration, and to change the code so that any exception would not affect subsequent timer callbacks—those would resume normally on the next animation frame. That way, an exception could cause localized damage, but not permanently break d3.timer or selection.transition.

natevw:
Ah, good point with the exception handling (at least [in V8](https://code.google.com/p/v8/issues/detail?id=1065#c3), "functions containing try/catch are not optimized").

I think what you're saying about letting exceptions be thrown naturally, is the same as I meant by "overhaul to be more robust". That does seem the way to go, then, though I can't really volunteer myself — after diving in this far, I decided the right fix was a trivial change to stop a particular bit of app code from throwing in the first place ;-)

mbostock:
I filed d3/d3-timer#4 to track this for 4.0, but I’m guessing we probably won’t make this proposed change to 3.x. I’ll leave open #2415 for now.

PR #2418, opened 2015-05-06
Add parens for d3_Map.

sghall:
I would go with the convention here and put the parens.  The result is the same, but I think the different style is confusing when looking across the code base.  In all other cases the parens are used when creating a map.  At first I thought maybe this was slightly more efficient or something.


mbostock:
I appreciate that some may disagree, but this it is our local style convention to omit the parens with no-argument constructors.

PR #2420, opened 2015-05-06
Layout.pie: Division by zero when sum(values) is 0

patriciaborges:
When d3.sum(values) == 0, k should be 0


mbostock:
Merged into #2591 for 3.5.7.

PR #2429, opened 2015-05-11
Add column function to dsv.parse

pietersv:
This adds an optional function to transform column names to `dsv.parse(...)`, comparable to the row accessor added in  https://github.com/mbostock/d3/pull/1107, 

The motivation is that input datafiles may have column names...
* with arbitrary case generated from other packages that are case insensitive like SAS or SPSS, requiring conversion to some defined case in Javascript
* with duplicate / missing column names, that need to be deduped, cleaned up, or invented, (e.g. see rules for "Variable Names" in [Stat/Transfer](https://www.stattransfer.com/support/manual/)

Tests are added that cover these two examples:
```js
var rows = d3.csv.parse(csv, null, function(name) { return name.toLocaleLowerCase(); })
var rows = d3.csv.parse(csv, null, function(name, i) { return "var"+i; })
```
http://stackoverflow.com/questions/26998476/right-way-to-modify-d3-csv-to-lower-case-column-names

### DIFF
--- a/src/dsv/dsv.js
+++ b/src/dsv/dsv.js
@@ -28,12 +28,12 @@ d3.dsv = function(delimiter, mimeType) {
     };
   }
 
-  dsv.parse = function(text, f) {
+  dsv.parse = function(text, f, n) {
     var o;
     return dsv.parseRows(text, function(row, i) {
       if (o) return o(row, i - 1);
       var a = new Function("d", "return {" + row.map(function(name, i) {
-        return JSON.stringify(name) + ": d[" + i + "]";
+        return JSON.stringify(n ? n(name, i) : name) + ": d[" + i + "]";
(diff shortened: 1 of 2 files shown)


mbostock:
I feel the existing row accessor function is sufficient for renaming columns. For example, to lower-case column names:

```js
var rows = d3.csv.parse(csv, function(d) {
  return {
    a: d.A,
    b: d.B,
    c: d.C
  };
});
```

It’s easiest if you know the column names, but you could, I suppose, write the code generically (at the cost of performance and complexity). Or you could replicate the approach taken by dsv.parse and wrap dsv.parseRows.

I agree that having a dedicated function for remapping columns is simpler, but I think the existing API is sufficient and we should endeavor to keep the API small.

PR #2443, opened 2015-05-20
octree and 3d force-directed graph layout

kalenedrael:
An octree and 3d force layout, almost entirely copied off of the existing quadtree and 2d force layout implementations with an extra dimension added. Appropriate tests have been added for the quadtree, but not the force layout.

If it's not appropriate for the base library, perhaps it could be a plugin?


mbostock:
Thanks for the pull request. I think a plugin does sound more appropriate given than D3 is primarily focused on 2D visualization.

amcdnl:
@kalenedrael did you break this out into its own plugin anywhere?

trusktr:
Live demo?

trusktr:
@kalenedrael Can you please share a link to a plugin version of this? Would be so awesome! :D

PR #2444, opened 2015-05-20
replaced utf8 greek letter with ascii names

danielbaak:
When using the non-minified version of d3, chromium Version 41.0.2272.76 Built on Ubuntu 14.04, running on LinuxMint 17.1 (64-bit) threw errrors on any code using non-ascii utf8 characters (e.g. greek letters).  For my own use in development I replaced all greek letters by their spelled-out ascii names.

I know that utf8 should be supported but for some reason in my case it wasn't.

All tests except test/geo/circle-test.js passed immediately. circle-test passed with slightly reduced floating point precision (1e-4 instead of 1e-5).  Circle-test didn't pass on my machine for the original HEAD commit either.

Cheers,
Daniel

PS: thanks for this awesome library. It make's visualizations (and more) fun.


mbostock:
Somewhat humorously, this automatic find-and-replace in this pull request breaks SI-prefix formatting (the `s` type in [d3.format](https://github.com/d3/d3-format#locale_format)), because µ is the SI prefix for micro, 10⁻⁶. At any rate, there’s no way to merge this pull request now that the code has been broken into [separate modules](https://github.com/d3) for D3 4.0. I do plan on replacing the Greek variable names with ASCII, however; see d3/d3-geo#8.

PR #2473, opened 2015-06-19
Feat/pack layout padding

hughes:
The padding parameter previously was constant across the layout, and required a second pass of d3_layout_packSiblings. Padding also would not apply to group nodes, only to the leaf nodes they contain.

With this change, it is possible to specify padding on any node, even group nodes. This value is added to the outer radius of the node when packing the nodes. This gives complete control over the spacing of any part of the layout. Since the computed padding is also exposed on the node via `node.p`, it is also possible to use the padding directly in the diagram. Padding is also computed on the first packing pass.

### Related issues

This addresses #2247 and #1686 but differs from the solution proposed in #2379, which adds consideration for the `value` property on a group node. 

### Padding vs value

The `value` of a node in pack layouts is treated as the relative *area* of a node. The `padding` of a node is added directly to the resulting *radius* of a node.

### Demos

[Demo](http://bl.ocks.org/hughes/aeee6c44713598563393)

[Demo 2](http://bl.ocks.org/hughes/9a5a68c0d49e02803e06)

### Tests

This change affected all pack layout tests, as the computed padding is added to the projection of the test results, which is zero in all previously built tests. Six additional tests have been added to ensure expected behaviour of padding constants and functions.


mbostock:
Related d3/d3-hierarchy#7.

There’s a proposed solution to that in d3/d3-hierarchy#8 now, so I want to compare these solutions.

mbostock:
This has been implemented in 4.0.

Issue #2055, opened 2014-10-09
Parsing of time offsets


jasondavies:
As per [ISO 8601](http://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC), in addition to ±hhmm we should support:

* Z
* ±hh:mm
* ±hh

*Edit: we are already compliant with strftime’s `%z`, which means ±hhmm, so it’s not clear whether we really need to support parsing the ISO 8601 variants.*

Mr0grog:
FWIW, I recently hit this problem when an upgrade in a client’s system changed their date output from using `Z` to using `+hhmm` for time zones. This was an side effect and unintentional, but they also didn’t expect it to be an issue since both were valid ISO 8601 strings.

For the same reason, I’d expected using `%Z` in the format specifier to solve the issue, but that doesn't handle `Z` style date strings. So if I want to be compatible with both versions, my date parsing becomes much more complicated and I have to have fallbacks. Since both `Z` and `+hhmm` are pretty common, it seems like making `%Z` able to parse both would be hugely valuable (and make things more safe at the same time). I don’t think formatting would need to change, just parsing.

I’d be happy to write a patch for it if you need somebody to take on this issue.

GaryRogers:
Yep, just hit this issue when trying to parse `2015-06-24T15:27:09-05:00` (produced by Moment.js as an ISO date) with the parse string `%Y-%m-%dT%H:%M:%S%Z`. Supporting `-05:00` in `%Z` would be grand.

mbostock:
Fixed in d3/d3-time-format#4 for 4.0.

PR #2480, opened 2015-06-26
Modify `%Z` time format specifier to parse any ISO timezone format

Mr0grog:
This addresses #2055. Since someone else also ran into this issue, I thought I’d submit a PR with the patch I wrote a little while ago. I’d love to see this go in, but I also realize there’s been zero discussion from any maintainers on the issue report for the last several months. Maybe this can spur some discussion?


mbostock:
Fixed in d3-time-format for 4.0.

PR #2481, opened 2015-06-28
Flexible class additional 

flowercat07:
Flexible class additional 

1. Check node class name
2. add class name


mbostock:
Closing in favor of the existing `selection.classed(name, true)` and `selection.classed(name, false)` for adding and removing.

Issue #1529, opened 2013-09-19
sub-second resolution axis ticks


bettse:
I'm working on a project that involves times that may be over a period of as little as a few seconds, and d3 seems to have a gap in its ability to render axis labels below second resolution.

Someone posted a question about this to [stack overflow](http://stackoverflow.com/questions/18461818/d3-js-time-scale-with-ticks-in-milliseconds), but most of the answers they got were about formatting, not about where the ticks were rendered at.  It did include [a jsFiddle that demonstates the problem](http://jsfiddle.net/SGyQN/).

The heart of the problem, from what I can tell, is that although javascript Date objects have resolution down to the millisecond, their representation often doesn't include this.  So far, the solution I've found that solves this best is to modify the d3_identity function and add an exception for Date:

``` javascript
function d3_identity(d) {
    if (d instanceof Date) {
        return d.getTime();
    }
    return d;
}
```

My worry is that this might be a special snowflake, and an incorrect way of solving the problem.  I was hoping to get some feedback regarding this solution, or where I should head in the source to produce a more elegant solution.

PR #2490, opened 2015-07-09
add d3.time.milliseconds to support millisecond resolution ticks

searler:
This addition makes it very easy to have millisecond resolution ticks on the axis 

e.g.
```
maxBottomAxis.ticks(d3.time.milliseconds, 100).tickFormat(d3.time.format('%L'));
```
Issue #1529 does not provide quite the same functionality since it requires an explicit modification of the data.

```
d.time = new Date(+d.time * 1000);
```

### DIFF
--- a/src/time/millisecond.js
+++ b/src/time/millisecond.js
@@ -1,0 +1,13 @@
+import "interval";
+import "time";
+
+d3_time.millisecond = d3_time_interval(function(date) {
+  return new d3_date(date);
+}, function(date, offset) {
+  date.setTime(date.getTime() + offset); // DST breaks setSeconds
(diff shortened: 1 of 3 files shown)


mbostock:
This is already supported in the new d3-scale module, which will be the basis of the 4.0 API; see [time.ticks](https://github.com/d3/d3-scale#time_ticks). For example:

```js
s.ticks("milliseconds", 100);
```

I’ve also added a [millisecond time interval](https://github.com/d3/d3-time#millisecond) to d3-time for completeness.

mbostock:
Closing since this is implemented in d3-hierarchy.

PR #2511, opened 2015-08-05
Setting attributes using a function.

greut:
Heya,

A little contribution to enable to set multiple attributes at once. Like `x` and `y` for a `svg` element.

Use case, being able to set two arguments at once (code from http://bl.ocks.org/mbostock/4207744)
```javascript
// Using translate
  svg.selectAll("text")
      .data(cantons.features)
    .enter().append("text")
      .attr("transform", function(d) { return "translate(" + path.centroid(d) + ")"; })
      .text(function(d) { return d.properties.name; });
// Using x/y
  svg.selectAll("text")
      .data(cantons.features)
    .enter().append("text")
      .attr(function(d) { var c = path.centroid(d); return {x: c[0], y: c[1]} })
      .text(function(d) { return d.properties.name; });
```

### DIFF
--- a/src/selection/attr.js
+++ b/src/selection/attr.js
@@ -13,6 +13,16 @@ d3_selectionPrototype.attr = function(name, value) {
           : node.getAttribute(name);
     }
 
+    // For attr(function), the function returns an object to be applied.
+    if (typeof name === "function") {
+        var n = name,
+            name = {};
+        this.each(function() {
+            var o = n.apply(this, arguments);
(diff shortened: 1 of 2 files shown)


mbostock:
This will be implemented in the [d3/d3-selection-multi module](https://github.com/d3/d3-selection-multi) for D3 4.0, so you should probably take a look over there.

greut:
Okay, thanks for the info!

PR #2516, opened 2015-08-07
fix bug of monotone curve in local extremum

xile611:
I find that the curve generated by d3_svg_lineMonotone sometimes has a “bump”, so i try to read the understand the algorithm in  <http://en.wikipedia.org/wiki/Monotone_cubic_interpolation> 。


mbostock:
Thank you for this pull request. The line and area interpolators (now called “curves”) are being completely re-written for 4.0; see [d3-shape](https://github.com/d3/d3-shape). @jviide has contributed a lovely new cubic monotone implementation based on [A Simple Method for Monotonic Interpolation in One Dimension](http://adsabs.harvard.edu/full/1990A%26A...239..443S) by M. Steffen.

Issue #2524, opened 2015-08-17
d3.range produces extra array element for fractional strides


andreasplesch:
The d3.range function produces an extra element as the last in the array for some fractional strides. For example:
```
    > d3.range(0,1,1/18).length
    < 19
```
is unexpected while
```
   > d3.range(0,1,1/17).length
   < 17
```
is expected.
This is perhaps due to floating point precision epsilons not being taken into account, or loss of precision at some type conversion.

andreasplesch:
I tried to see what happens in array/range.js if step is 1/18. The k upscale factor will be 1e17.
So the upscaled step size will be 5555555555555555 . But floating point precision loss leads to
```
    > 18 * 5555555555555555
    < 99999999999999980
```
which is smaller than the upscale stop value of 1e17, and leads to the extra element.
(so I am not sure what is gained by the upscaling ?)
A fix may be to adjust stop in line 16 with something like:
```
stop = (stop - 1e-16) * k;
```
although this may need to be reversed for negative steps.
Or to not upscale and more carefully use the smaller than and larger than comparisons in the while loops.

andreasplesch:
As a workaround  without fixing the range function what works in my case is:
```
var eps = 1e-16
d3.range(0,1-eps,1/18).length
18
```
I empirically tested this workaround by
```
d3.range(1,10000).forEach( function(d) {
  if (d3.range(0, 1-eps, 1/d).length !== d) {console.log(d) ;}})
```
which does not find any unexpected range for eps = 1e-16 but does find many for eps = 1e-17;
This is equivalent to the suggestion above though it is just for the [0,1[ domain.

andreasplesch:
Ok. I noticed that the API documentation actually mentions that the result is less predictable with fractional step sizes. Now, I know what that means and the behaviour can be considered somewhat documented with this issue. Perhaps there is no good fix ?

andreasplesch:
d3/d3-arrays/src/range.js approach works much better. So I think this will become resolved in the next version of d3.

andreasplesch:
Hm, while the new range.js works better, a little bit deeper empirical testing finds the same issue there. See https://github.com/d3/d3-arrays/issues/5

PR #2526, opened 2015-08-18
make fractional steps more predictable

andreasplesch:
An attempt to take into account limited floating point precision with the goal of making the number of returned array elements more predictable. See issue #2524.

### DIFF
--- a/src/arrays/range.js
+++ b/src/arrays/range.js
@@ -12,8 +12,9 @@ d3.range = function(start, stop, step) {
   var range = [],
        k = d3_range_integerScale(abs(step)),
        i = -1,
-       j;
-  start *= k, stop *= k, step *= k;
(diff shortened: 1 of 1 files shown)


andreasplesch:
It took me a while to find the d3/d3-arrays project. The approach in range.js there works much better and makes the commit above obsolete. Perhaps it is useful as a temporary fix until the next major version of d3.

mbostock:
Thanks. Glad to hear this will be fixed in 4.0. Sorry for the confusion; it should be temporary.

andreasplesch:
Hm, it looks like I spoke to soon. While the new range.js works better, it still has the same issue. I opened a new, crossreferenced issue there: https://github.com/d3/d3-arrays/issues/5

PR #2528, opened 2015-08-21
Create sv-SE.js - Swedish locale

leitzler:
Added support for Swedish locale


mbostock:
I think you want `"\xa0"` here instead of `"\0xa0"`; the latter is the character NULL followed by “xa0”. I can merge and then fix.

PR #2530, opened 2015-08-23
Generic getter-setter property.

linhphi9x94:


mbostock:
Thanks for the pull request. I think it’s probably not worth doing this. It might be a tiny improvement here, but it’d be a fair amount of work to change all of the code to adopt this pattern. And moreover, this pattern doesn’t allow you to have custom code for side-effects or validation when a property is set, so you’d end up with some properties than can use the generic getter-setter and others written the “long” way because they need to do something special.

PR #2536, opened 2015-08-27
Add margin parameter to pack layout

larskotthoff:
This PR adds the `margin` parameter to the pack layout. It allows to specify the distance to the parent circle in the same way that the `padding` parameter allows to specify the distance to the sibling circles.

Demo [here](http://bl.ocks.org/larskotthoff/085bb57f710aa417af41). I implemented this in response to [a question on Stackoverflow](https://stackoverflow.com/questions/32235067/padding-and-margin-on-pack-layout).


mbostock:
Moved to d3/d3-hierarchy#7 for 4.0.

PR #2537, opened 2015-08-27
Add missing import

rubenv:
Without this, custom builds that depend on behavior/zoom, yet don't
include transitions fail to include the needed d3_selection_interrupt.

### DIFF
--- a/src/behavior/zoom.js
+++ b/src/behavior/zoom.js
@@ -4,8 +4,8 @@ import "../event/drag";
 import "../event/event";
 import "../event/mouse";
 import "../event/touches";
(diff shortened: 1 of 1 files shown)


rubenv:
No wait, this doesn't work, d3_selection_interrupt depends on d3_transitionNamespace.

Should we add a dependency on the whole of `transition` from inside `behavior/zoom`?

mbostock:
Thanks for the pull request. SMASH is going away in 4.0 with the adoption of Rollup, so it’s probably not worth further effort to fix this in 3.x; Rollup uses ES6 imports and will be much more robust.

PR #2547, opened 2015-09-07
Pointer (and MSPointer) events for the zoom behavior.

lubomirov:
Added support of Pointer events in addition to Touch events


mbostock:
Fixes #1439. Related #2403.

nikoloas:
Hi, I have just downloaded this version and the pinch zoom/panning is not working for my d3 SVG. Do I need to implement something different to add support for IE11?

lubomirov:
Hi, nikolas, I've made rebase and double-checked my patch suggested here. It works in IE11 (particularly in 11.0.25). With this version you can do zoom with two/three/more fingers (there is an algorithm that considers third finger if second one goes out of zoom area). Thank you for downloading and reminding me about this pull request.

nikoloas:
Many thanks for the heads up, I will download this version and test it tomorrow.

Greyjoy:
Hi, I'm using d3 to draw and manipulate a map and it works fine, excepted on windows mobile devices (pinch zoom and translations by touch are disabled). Maybe those changes could fix it ? Can it be merged with main project ? 
Many thanks for your work.

Leo.

PR #2548, opened 2015-09-09
Fix drag behavior on for <svg:use> elements

stefwalter:
Drag behavior is broken for <svg:use> elements on certain
browsers. This includes ```d3.layout.force(...).drag()```

On browsers like Internet Explorer 11, the ```<svg:use>```
EventTarget is represented by a SVGElementInstance that looks
like this:

```
    [object SVGElementInstance]
       {
          [functions]: ,
          __proto__: { },
          childNodes: { },
          constructor: { },
          correspondingElement: { },
          correspondingUseElement: { },
          firstChild: null,
          lastChild: null,
          nextSibling: null,
          parentNode: { },
          previousSibling: { }
       }
```

No other properties such as ```event.target.style``` are present. We must use the
```event.target.correspondingElement``` property to access the actual SVG element
instantiated by the ```<svg:use>``` in order to have a successful drag.


stefwalter:
@abonas @jwforres FYI

stefwalter:
To repdroduce this bug:

 * Use an Internet Explorer appliance from here. IE11 on Win7
 * Clone https://github.com/kubernetes-ui/topology-graph
 * ```grunt depends```
 * ```grunt serve --hostname=0.0.0.0```
 * Access port 9000 of your host with IE

stefwalter:
Rebased

jwforres:
@mbostock this is a relatively small change, what are the chances on getting it merged?

jwforres:
Also i'm pretty sure this was an issue with Safari, not just IE

stefwalter:
@jwforres Above are relatively simple instructions to reproduce this bug. If you have access to Safari, you can try them out.

jwforres:
@spadgett has faster access to safari than I do :)

mbostock:
This fix is trivial, which I like, so this seems totally reasonable. I’ll try to get it out this week. I’m mostly focused on D3 4.0 but this seems like a safe and tiny change. Thank you for the fix.

mbostock:
Released in 3.5.15. Thank you!

jwforres:
Thanks for merging!

mbostock:
It’d probably be better to fix this in [`d3_event_dragSuppress`](https://github.com/mbostock/d3/blob/master/src/event/drag.js) rather than in the drag behavior, since it’s only there where the code depends on the event target having a style property. I’ll try to do that for 4.0.

PR #2550, opened 2015-09-09
Only prevent default on mousewheel event when trying to zoom within the scaleExtent

jakerichan:
fixes https://github.com/mbostock/d3/issues/2549


jakerichan:
@mbostock I notice you closed the issue relating to this PR.  Is there anything I need to update before we merge?

mbostock:
The only reason I closed the issue is that this pull request is sufficient to track the request.

Concerns with this change:

* A mousewheel event could initiate a change from an allowed scale extent to a disallowed scale extent, and in that case, we want to clamp the new scale extent to the minimum or maximum allowed value and still prevent default on the extent. (Your implementation appears to only check the new value, and not whether it is different from the current value.) For example, if the scale is clamped in [1, 10], and a mousewheel would change the scale from 8 to 12, then we want to change it to 10 and prevent default.

* I’m not totally sure the logic is right. For example, this still sets the mousewheelTimer even when the mousewheel event is going to be ignored. The zoom behavior event handling logic is complicated, so it’ll require some thought.

* I haven’t decided whether the zoom behavior should always behave this way, or whether it should be something that the user can enable. Probably it would be reasonable to just make it always behave this way, and consider it a bug fix.

mbostock:
This has been implemented in [d3-zoom](https://github.com/d3/d3-zoom) for D3 4.0.

PR #2552, opened 2015-09-11
Check if NaN

jamesleesaunders:
Check if resulting endAngle is NaN, if it is set it to 0. Fixes [Issue 2475](https://github.com/mbostock/d3/issues/2475)


mbostock:
Looks like #2420 is probably a better fix.

Issue #2554, opened 2015-09-14
Preserve comments in distributed d3.js single file


JabbyPanda:
I would be nice to alter Makefile script to include `--comments` [uglify] (https://github.com/mishoo/UglifyJS2/blob/master/README.md) library directive that will include comments into resulted single `d3.js` file in dist folder. 

I guess only developers, not end-users are interested in`d3.js` file and developers like commented code.

mbostock:
Will respond in #2555.

PR #2555, opened 2015-09-14
Uglify Preserve Comments

jamesleesaunders:
Addresses suggestion: Preserve comments in d3.js (issue #2554)
https://github.com/mbostock/d3/issues/2554

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ d3.zip: LICENSE d3.js d3.min.js
 
 d3.js: $(shell node_modules/.bin/smash --ignore-missing --list src/d3.js) package.json
 	@rm -f $@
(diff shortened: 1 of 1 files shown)


mbostock:
This adds the `--comments` argument after the `-o` argument, causing an error. But more importantly, the `--comments` argument isn’t passed to UglifyJS because of our custom wrapper script. The correct fix is probably something like this in bin/uglify:

```diff
-    output = uglify.OutputStream({ascii_only: true}),
+    output = uglify.OutputStream({ascii_only: true, comments: true}),
```

That didn’t seem to do anything, though, and I’m not going to investigate it any further at the moment. Because…

D3 4.0 has a new build process using [Rollup](http://rollupjs.org/). This preserves comments and formatting (for the most part) in the generated file.

PR #2567, opened 2015-09-30
make d3.map / d3.set more like ES6 Map / Set

vp2177:
Although .size is still a function and not a getter as in ES6

### DIFF
--- a/src/arrays/map.js
+++ b/src/arrays/map.js
@@ -20,8 +20,6 @@ function d3_Map() {
   this._ = Object.create(null);
 }
 
-d3_Map.length = 0;
-
 var d3_map_proto = "__proto__",
     d3_map_zero = "\0";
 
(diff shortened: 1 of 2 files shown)


mbostock:
Adding map.clear and set.clear methods seems reasonable, but I think it’s a losing battle to try to make these closer to ES6. For example: D3’s map.set returns the set value, not the map instance; D3’s map.forEach passes arguments {key, value, map} rather than {value, key, map}. Also, map keys and set values are coerced to strings.

I want these classes to go away in the future, when it’s reasonable for D3 to only support ES6 runtimes. I don’t think we’re ready for that yet, and because of the nature of the ES6 Map and Set APIs, polyfilling is not a great transitional solution.

mbostock:
If we do modify the map & set API, I’d prefer to do it in d3-arrays for 4.0 rather than for 3.x.

PR #2568, opened 2015-09-30
make more like ES6 Set

vp2177:
Even though in a sense the ES6 spec isn't intuitive (.keys() / .entries()?)


vp2177:
merged into :patch-2

Issue #2468, opened 2015-06-11
Stopping and starting the force layout causes duplicate timers.


mbostock:
Mentioned in #1790:

> Currently, if you .stop() a force layout and restart it before force.tick() is called, you end up with two timers, both hitting the same call-back and the animation runs at double speed.

mbostock:
Fixed in #2591 for 3.5.7.

PR #2600, opened 2015-10-25
Cancel pre-empted transitions synchronously.

mbostock:
When a scheduled transition is pre-empted by a newer transition starting, we now cancel the old transition’s timers immediately, rather than waiting for the old transition to start.

TODO:
- [x] d3.timer should return void
- [x] verify whether those errors can be thrown


mbostock:
Related #2377 #2189.

mbostock:
Related #2468, in that the new internal d3_timer API should provide a mechanism for the force layout to cancel its timer and avoid starting it twice.

Issue #2589, opened 2015-10-19
Mouse wheel implementation broken on Windows


blaise-io:
In Chrome 46 / Windows 7, `d3.event.deltaY` always returns `-0`, so mousewheel does not do anything. I can fix it by replacing:

```
    if (!d3_behavior_zoomWheel) {
      d3_behavior_zoomWheel = "onwheel" in d3_document ? (d3_behavior_zoomDelta = function() {
        return -d3.event.deltaY * (d3.event.deltaMode ? 120 : 1);
      }, "wheel") : ...
```

with:

```
    if (!d3_behavior_zoomWheel) {
      d3_behavior_zoomWheel = "onwheel" in d3_document && !(/win/i).test(navigator.platform) ? (d3_behavior_zoomDelta = function() {
        return -d3.event.deltaY * (d3.event.deltaMode ? 120 : 1);
      }, "wheel") : ...
```

See: https://github.com/layar/d3/commit/49077acd0c847f9dc9811efb96d963e323d32a54, tested in latest Chrome, IE, Firefox on Windows and Mac OS X.

This works, but it would be better to use a proven mouse wheel library to handle mouse wheel events, or draw inspiration from https://developer.mozilla.org/en-US/docs/Web/Events/wheel#Listening_to_this_event_across_browser

langdonx:
FWIW, on two separate machines Chrome 46+Windows 7, deltaY is either -100 or +100 and zoom still works fine.

kfei:
Possibly related to Chrome's breaking changes: https://codereview.chromium.org/1285133008/

Issue #1048, opened 2013-02-01
Hidden div for scroll wheel events not being used


tmcw:
Example: http://bl.ocks.org/4692489

(refs #1047)
This uses d3 master with the try/catch block for the hidden div removed, and just the hidden-div condition:

``` diff
diff --git a/d3.js b/d3.js
index dad2c73..1a8f619 100644
--- a/d3.js
+++ b/d3.js
@@ -3904,13 +3904,9 @@
       d3_behavior_zoomDiv = d3.select("body").append("div").style("visibil
     }
     var e = d3.event, delta;
-    try {
-      d3_behavior_zoomDiv.scrollTop = 1e3;
-      d3_behavior_zoomDiv.dispatchEvent(e);
-      delta = 1e3 - d3_behavior_zoomDiv.scrollTop;
-    } catch (error) {
-      delta = e.wheelDelta || -e.detail * 5;
-    }
+    d3_behavior_zoomDiv.scrollTop = 1e3;
+    d3_behavior_zoomDiv.dispatchEvent(e);
+    delta = 1e3 - d3_behavior_zoomDiv.scrollTop;
     return delta;
   }
```

On my machine (MBP, OSX 10.8.2, USB mouse), I get

```
Uncaught Error: DISPATCH_REQUEST_ERR: DOM Events Exception 1 
```

in Chrome 25.0.1364.58 beta

```
NS_ERROR_ILLEGAL_VALUE: Illegal value
```

in Firefox 20.0a2

```
Uncaught exception: Error: UNSPECIFIED_EVENT_TYPE_ERR
Error thrown at line 1, column 43739 in Ve() in http://localhost:3000/d3.min.js:
    return Ja.scrollTop=1e3,Ja.dispatchEvent(n),t=1e3-Ja.scrollTop
called from line 3, column 60881 in c() in http://localhost:3000/d3.min.js:
    d||(d=n(d3.mouse(this))),r(Math.pow(2,.002*Ve())*x),i(d3.mouse(this),d),a(w.of(this,arguments))
called via Function.prototype.apply() from line 1, column 15818 in i(t) in http://localhost:3000/d3.min.js:
    n.apply(a,o)
```

In Opera 12.13

PR #2605, opened 2015-11-04
Add a property for a custom zoom factor to the zoom behavior.

eseifert:
When using the zoom behavior at the moment zooming by mouse wheel or double click causes a zoom which is a multiple of 2. A custom zoom factor would allow more control over zooming speed, especially with mouse wheels.

### DIFF
--- a/src/behavior/zoom.js
+++ b/src/behavior/zoom.js
@@ -17,6 +17,7 @@ d3.behavior.zoom = function() {
       scaleExtent = d3_behavior_zoomInfinity,
       duration = 250,
       zooming = 0,
+      zoomFactor = 2,
       mousedown = "mousedown.zoom",
       mousemove = "mousemove.zoom",
       mouseup = "mouseup.zoom",
@@ -138,6 +139,12 @@ d3.behavior.zoom = function() {
     return zoom;
   };
 
+  zoom.zoomFactor = function(z) {
+    if (!arguments.length) return zoomFactor;
+    zoomFactor = z;
+    return zoom;
+  };
+
   function location(p) {
     return [(p[0] - view.x) / view.k, (p[1] - view.y) / view.k];
   }
@@ -159,7 +166,7 @@ d3.behavior.zoom = function() {
   function zoomTo(that, p, l, k) {
     that.__chart__ = {x: view.x, y: view.y, k: view.k};
 
-    scaleTo(Math.pow(2, k));
+    scaleTo(Math.pow(zoomFactor, k));
     translateTo(center0 = p, l);
 
     that = d3.select(that);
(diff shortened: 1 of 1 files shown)


mbostock:
I worry that a better fix would be to make mousewheel handling more reliable across browsers.

In the past we went to elaborate measures to ensure consistent behavior by redispatching the mousewheel event to an invisible scroll area, and then measuring how much the scroll offset changed. (Scrolling is much more consistent than mousewheel events.) We later removed that hack because it was, as you can imagine, quite elaborate just to receive a correct delta amount, and it appeared that most modern browsers were now better behaved. See #1048.

But judging from the issues I’m seeing, maybe mousewheel is broken again, at least on some platforms?

Related #2605 #2589 #2022.

Aside from that, a quick observation on this implementation: it breaks the intended behavior of dblclick snapping to round zoom levels. This is less important these days with the increased prevalence of vector map tiles but it’s a big deal if you’re using the zoom behavior with raster tiles (such as [bl.ocks.org/4132797](http://bl.ocks.org/mbostock/4132797)).

The idea is that if you use mousewheel to go to a fractional zoom (by which I mean the scale is an integer power of the zoom factor, so a zoom of 1.5 is a scale of 2^1.5 = 2.8284…) then a dblclick would take you to the smallest greater round zoom (2) and a shift-dblclick takes you to the largest lesser round zoom (1). If you are already on a round zoom level, then it just increments or decrements the zoom level as appropriate.

eseifert:
Thanks Mike! Actually, the changes aren't intended as a fix to mouse wheel handling.

My use case is a story telling application where I use the zoom level to show or hide elements. i want to be able to define a certain number of zooming steps between minimum and maximum scaling for the mouse wheel or double click.

At the moment it's hard to change the implemented base-of-2 zoom. I tried to create a modified version of D3's zoom behavior, but i just ended up pulling a lot of internals into my code.

You're right, I broke the dblclick snapping. I guess i should fix that. I changed it because it didn't work as intended with non-base-2 zoom for my use case: With `zoomFactor=1.5` and `scale=0.8`, zooming out with dblclick would result in a scale of 0.8 as a result of rounding errors.

Coming to mouse wheel interaction: What particular reason is there for the .002 factor in zoom behavior calculation?

I did some research on cross-browser mouse wheel handling and I think it will be very hard to get right. All of the existing solutions I've looked at have issues, like *React*'s solution (https://github.com/facebook/fixed-data-table/blob/d6bac3ccc59612d521b29d12d84f5142e5f252cb/dist/fixed-data-table.js#L1947-2113) or the *jQuery mousewheel plugin* (https://github.com/jquery/jquery-mousewheel/blob/master/jquery.mousewheel.js). My temporary solution is to use `Math.sign(wheelDelta)` instead of `wheelDelta*.002`.

eseifert:
I reverted the changes to dblclick snapping, so it's just about the zoomFactor now.

PR #2608, opened 2015-11-06
fix collections `forEach` method to follow JS spec

nicksrandall:
In order to be consistent with the Map and Set spec, I have refactored
the forEach method on both collection primitives.

**THIS DOES INTRODUCE A BREAKING CHANGE**


nicksrandall:
see https://github.com/d3/d3-arrays/issues/12

nicksrandall:
Issue also reported here: https://github.com/mbostock/d3/issues/1413#issuecomment-21328564

mbostock:
We can’t make this change in 3.x, and d3-arrays would be to make this change for 4.0+, so closing this pull request. Will respond over in d3/d3-arrays#12.

Issue #1005, opened 2013-01-10
Do the zoom and drag behaviors need to cancel click events?


ansis:
https://github.com/mbostock/d3/blob/master/src/behavior/zoom.js#L113-L118
It cancels the first click event after the drag, the one assumed to be created by the drag. Is there any reason it needs to be cancelled? Could leaving the click event cause trouble?

It also seems to cause problems with Opera, which doesn't trigger click events for drags, which could be an Opera bug or an accepted inconsistency (not sure yet). Ran into that at systemed/iD#357

airandfingers:
I've observed inconsistent behavior with v3.5.8 in Chrome, as have others in the comments at http://stackoverflow.com/a/19951105/877682.

I don't see this behavior with simple examples like http://jsfiddle.net/2EqA3/3, leading me to believe that the inconsistency has to do with our draggable nodes, which are complex, comprising multiple path and text elements

I've worked around the problem by changing d3: when dragRestore doesn't check if d3.event.target === target, this issue goes away for me. I haven't observed side effects caused by this, but I don't understand the code well enough to know what to test. https://github.com/mbostock/d3/commit/a8de36bd85cecafbc76f6e4c7e5d80b9b280b500

mbostock:
The issue you are seeing is indeed related to draggable nodes with multiple elements. See #312 and specifically @jasondavies’ comment:

> The problem is that the browser only fires a click event if mousedown and mouseup occur on the same element. So if you have a mousedown occurring on one element, and then mouseup on another, no click event is fired. This could happen in a number of ways e.g. if you clamped the dragged element and mouseup occurred outside the clamp region.
> 
> We had mistakenly assumed that a click event is always fired after mousedown + mouseup, so we always cancelled the next click event. In the original bug report, the OP found that he had to double-click on links after dragging a different element, because the next click was always being cancelled. We only want to cancel the click if it is actually going to be fired as a result of the mousedown + mouseup combination, which isn't always the case.

That said, it looks like we don’t need this sanity-check since we added the 0ms timeout, which automatically expires click suppression if a click event doesn’t immediately fire. In other words, since we only suppress an *immediate* click event, we don’t have to worry about false positives due to inconsistent browser behavior. (Before the 0ms-timeout, the click could be suppressed indefinitely!)

So, I’d be okay with removing target checks and relying on the timeout for click suppression.

airandfingers:
Great, thanks for the context and fix!

PR #2616, opened 2015-11-11
Suitably zealous click canceling.

mbostock:
This reverts #312, since we later implemented a more robust check in #1005 using a zero-delay timeout to expire click suppression.


mbostock:
Merged into #2638 for 3.5.9.

JesperWe:
Was this supposed to fix the defaultPrevented problem on Chrome? I use a zoom behaviour with drag for panning, and need to recognize clicks as well. I am on 3.5.16 now, and I still have issues with Chrome always setting defaultPrevented = true, both on drags and clicks, which means I can't tell them apart. Firefox works fine, sets defaultPrevented = false for clicks.

mbostock:
Works for me? ¯\\\_(ツ)\_/¯

http://bl.ocks.org/mbostock/a84aeb78fea81e1ad806

JesperWe:
OK, my case is different: I have a zoom behavior, where a drag does panning, but a click should do something else. In this case both a single click and a drag event triggers a 'click', which comes with `defaultPrevented=true` for **both** types of operation on Chrome, but `defaultPrevented=false` for the single click and `defaultPrevented=true` for the drag on Firefox and Edge. Maybe this is a different issue?

mbostock:
Still works for me with zoom? ¯\\\_(ツ)\_/¯

http://bl.ocks.org/mbostock/35964711079355050ff1

JesperWe:
Right. I have spent the day investigating this. Before this commit, my app was intermittently buggy. With this patch it stopped being intermittent, so it was easier to debug. I had the `.on( 'click' )` on the `<svg>`, and the `.call(zoomBehavior)` on a `<g>` tag inside it. With this setup there is a difference between Chrome and other browsers. When I moved the click to the same tag as the zoom, it starts to work as expected.

ljani:
I just tried the [Zoom vs. Drag](http://bl.ocks.org/mbostock/35964711079355050ff1) example and it doesn't work for me with Chrome 50.0.2661.102 m on Windows 10. I'm only able to zoom and pan.

Updating to Chrome 51.0.2704.63 m seems to fix the issue, but it's very new and not everyone has updated yet.

mbostock:
That sounds like the expected behavior for that example; I’ve just renamed it Zoom vs. Click to make the intent clearer. It is not intended to allow dragging of individual circles; it is intended to demonstrate how to allow clicks on a zoomable element.

Here’s a [zooming and dragging example](https://bl.ocks.org/mbostock/6123708) in D3 3.x.

Here’s a [zooming and dragging example](http://bl.ocks.org/mbostock/2b534b091d80a8de39219dd076b316cd) in D3 4.0.

ljani:
@mbostock I thought the background was supposed to flash if you clicked it? With Chrome, there are times when it won't flash even if I clicked anywhere. Most of the time it works just fine.

mbostock:
It only flashes if you click on the background. Furthermore, if you move the mouse at all between mousedown and mouseup, it won’t count as a click because it is interpreted as a (tiny) drag gesture. If you have a particularly sensitive mouse, it’s possible that you’re seeing a tiny bit of mouse movement just in a normal mousedown and mouseup. This is one reason not to try to support both clicking and zooming: it’s inherently ambiguous.

Right-clicking is probably an unrelated issue. The zoom behavior doesn’t ignore right clicks by default in 3.x; it does in 4.0.

ljani:
Yeah, I meant the background. And I agree with your thoughts about clicking and zooming.

Here's another thing: Tooltips. Consider [this example](http://bl.ocks.org/ljani/7d11c3fb9ab99776f95e3b861f61706b). If you:

1. Point to a circle so that it'll display a tooltip (= the `<title>`)
2. Move your cursor over the white background
3. Click it, nothing happens on Chrome. On Firefox, the background flashes.

ljani:
Here we go: https://gfycat.com/SolidFabulousHorsefly
Each time I click the background, `defaultPrevented` is logged to the console. I'm still not sure how to end up here, but it seems that leaving Chrome open and doing other stuff seems to be the key, heh.

PR #2650, opened 2015-11-24
Fix #2646 tree root handle one child

### LINKED ISSUE #2646: error occurs when tree's root node has only one child
Hi,

I'm using d3 to generate a radial tree, I found it can't be shown when the root node's children has only one item. The error message is like: 

Error: Invalid value for <path> attribute d="MNaN,NaNCNaN,NaN NaN,NaN NaN,NaN"a @ d3.v3.min.js:1
d3.v3.min.js:1 Error: Invalid value for <g> attribute transform="rotate(NaN)translate(0)"a @ d3.v3.min.js:1
d3.v3.min.js:1 Error: Invalid value for <g> attribute transform="rotate(NaN)translate(360)"

see here for detail: http://plnkr.co/edit/0fhh6B?p=preview

KaranLala:
Hey Mike! I added  fix for #2646. 
The issue was the tree.separation method was using the depth of the left and right nodes. In the case in which the root had only one child, both nodes were equal to root (left and right = root) which has a depth of 0 causing it to fail. So, on line 7053 of d3.js I made if (node.x < left.x) left = node; to if (node.x <= left.x) left = node;
I hope that is an appropriate fix. It passed all the tests.
Also, thank you for making d3. It is awesome!

### DIFF
--- a/d3.js
+++ b/d3.js
@@ -7050,7 +7050,7 @@
       if (nodeSize) d3_layout_hierarchyVisitBefore(root0, sizeNode); else {
         var left = root0, right = root0, bottom = root0;
         d3_layout_hierarchyVisitBefore(root0, function(node) {
-          if (node.x < left.x) left = node;
+          if (node.x <= left.x) left = node;
           if (node.x > right.x) right = node;
           if (node.depth > bottom.depth) bottom = node;
         });


vylan:
Thanks KaranLala, for this case it's a nice fix. But I found in the case of the root node having no child, error still occurs, could you provide an improved fix, thanks in advance. (http://plnkr.co/edit/XFZSeY?p=preview)

KaranLala:
Hey vylan, I looked into it further and my fix is only useful for this case. I'll try to find a better solution to the bigger problem (root with no child included). For now, this quick fix in the tree.separation function should help avoid the issue - 
function(a, b) { 
    	if (!a.depth) return 1;
    	return (a.parent == b.parent ? 1 : 2) / a.depth;
}

PR #2654, opened 2015-11-25
Fix bug in force in which simulation did not stop if alpha was set to…

linuss:
… zero

### DIFF
--- a/src/layout/force.js
+++ b/src/layout/force.js
@@ -201,12 +201,15 @@ d3.layout.force = function() {
       if (x > 0) { // we might keep it hot
         alpha = x;
       } else { // or we might stop
-        timer.c = null, timer.t = NaN, timer = null;
+        timer = null;
         event.start({type: "end", alpha: alpha = 0});
       }
     } else if (x > 0) { // otherwise, fire it up!
       event.start({type: "start", alpha: alpha = x});
       timer = d3_timer(force.tick);
+    } else { // otherwise, we're done
+      timer = null;
+      event.start({type: "end", alpha: alpha = 0});
     }
 
     return force;


mbostock:
Are you sure there’s a bug here? I don’t see it from just looking at your change. There are four cases to consider with this logic:

1. The simulation is active (alpha > 0) and the new alpha is positive (x > 0). The timer is already running, so just set alpha.
2. The simulation is active (alpha > 0) and the new alpha is not positive (x <= 0). The timer is running and we want to stop the simulation, so stop the timer.
3. The simulation is inactive (alpha = 0) and the new alpha is positive (x > 0). The timer is not running and we want to start the simulation, so start the timer.
4. The simulation is inactive (alpha = 0) and the new alpha is not positive (x <= 0). The timer is not running, so do nothing.

To illustrate in the code:

```js
if (alpha) { // Cases 1 & 2.
  if (x > 0) { // Case 1. Set alpha.
    alpha = x;
  } else { // Case 2. Stop the timer.
    timer.c = null, timer.t = NaN, timer = null;
    event.start({type: "end", alpha: alpha = 0});
  }
} else if (x > 0) { // Case 3. Start the timer.
  event.start({type: "start", alpha: alpha = x});
  timer = d3_timer(force.tick);
} else { // Case 4. Do nothing.
}
```

Also, for your fix, note that setting the `timer` to null doesn’t have the intended effect: it’s still in the timer queue, so it will still receive callbacks. It just means you’ve lost the reference to the timer. The way to stop the timer is to set its callback (`timer.c`) to null. (In 4.0, there will be a public API for this, [timer.stop](https://github.com/d3/d3-timer#timer_stop); for now, the force layout uses a private API.)

That said, there *is* a bug in that we’re dispatching a *start* event when the force layout ends manually (albeit with the event type of “end”), when we should be dispatching an *end* event. I’ll commit this fix:

```diff
--- a/src/layout/force.js
+++ b/src/layout/force.js
@@ -202,7 +202,7 @@ d3.layout.force = function() {
         alpha = x;
       } else { // or we might stop
         timer.c = null, timer.t = NaN, timer = null;
-        event.start({type: "end", alpha: alpha = 0});
+        event.end({type: "end", alpha: alpha = 0});
       }
     } else if (x > 0) { // otherwise, fire it up!
       event.start({type: "start", alpha: alpha = x});
```

linuss:
You're completely right, I'm sorry for my hasty PR. I encountered this bug (which in retrospect was probably just caused by the 'start' event), and wanted to help out.

PR #2658, opened 2015-11-27
npm ignore the `src` dir`

fishbar:
because `src/` is useless in the npm module

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -3,3 +3,4 @@ test/
 lib/
 .DS_Store
 _site
+src/


mbostock:
Eh, I don’t see much harm in leaving it in. In D3 4.0, the `src` folder will contain ES6 modules (rather than D3 3.x’s ES6-like [SMASH](https://github.com/mbostock/smash) files. There will still be a prebuilt UMD module for CommonJS and other non-ES6 environments, but it’s my plan to include the ES6 source for future compatibility and because many people use npm as a simple package manager outside of a strict CommonJS environment.

PR #2668, opened 2015-12-07
Tree with only one branch overlays the objects.

jfrickson:
The Y coordinate gets set to zero for all children/grandchildren, etc.

### DIFF
--- a/src/layout/tree.js
+++ b/src/layout/tree.js
@@ -28,8 +28,8 @@ d3.layout.tree = function() {
           bottom = root0;
       d3_layout_hierarchyVisitBefore(root0, function(node) {
         if (node.x < left.x) left = node;
-        if (node.x > right.x) right = node;
-        if (node.depth > bottom.depth) bottom = node;
+        if (node.x >= right.x) right = node;
+        if (node.depth >= bottom.depth) bottom = node;
       });
       var tx = separation(left, right) / 2 - left.x,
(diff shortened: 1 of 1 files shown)


jfrickson:
Sorry, that wasn't the problem

mbostock:
This looks like a duplicate of #2650 (which was retracted by the author… I haven’t had time to investigate why). Appears to be a fix for #2646, also.

jfrickson:
Ok, I also put in a fix for his and vylan's "root only node" problem.

vylan:
Thanks @jfrickson , I 've tested with your code. I believe it fixed the issue. by the way, please change seq2 to sep2.

jfrickson:
Should be good to go!

mbostock:
I’ve pushed a slightly simpler fix in d3/d3-hierarchy@68055b9d22b2b1f002fd2cc1995238cbb778f644 for 4.0. In the case of a single-branch tree, the left- and right-most node is always the root, and I think it’s best to avoid invoking the separation function when *a* and *b* are the same node (the root); in that case, we can simply center the tree within the layout area.

I’m probably not going to back-port the fix to 3.x. A simple workaround is to not return Infinity from the separation function:

```js
function separation(a, b) {
  return (a.parent === b.parent ? 1 : 2) / Math.max(1, a.depth);
}
```

PR #2675, opened 2015-12-11
Remove trailing 0 error from chord group value

zhoux10:
- (x - x0) / k sometimes provides a value with trailing 0's due to division. This means groups that have an integer value appear to be a decimal number

Example of problem below:
image

With changes:
image


mbostock:
The fix looks good.

(If your values weren’t numbers, I would say to use d3.format when displaying the sum, but this fix seems totally reasonable.)

PR #2679, opened 2015-12-14
Adding "firstTick" method to the axis function

jharveymerkle:
Sometimes there is not enough screen real estate to use the standard axis label for your axes. (This is especially true when packing in a lot of Tufte-style small multiples.) This code change adds a "firstTick" method to D3's axis function so that you can replace the first value in your axis with a custom text string. In the example block (linked below), I have replaced the first value of of the X-axis with the custom string "Year".

http://bl.ocks.org/jharveymerkle/4ca2bf750b02cb724cff

The files changed are src/svg/axis.js and test/svg/axis-test.js.

This update to D3 is offered with permission of Merkle, Inc. 
http://www.merkleinc.com


mbostock:
The recommended way of implementing this feature is post-selection, as in the [custom axis example](http://bl.ocks.org/mbostock/4323929). After calling the axis, select the first tick (or whatever else you want to customize) and then apply the desired changes. For example:

```js
svg.append("g")
    .call(axis)
  .select(".tick")
    .text("First!");
```

If you just want to modify the tick text, another way to implement this feature is with a custom tick format, and then use the tick value (`d`) or the index (`i`). For example:

```js
axis.tickFormat(function(d, i) { return i ? d.toFixed(2) : "First!"; });
```

PR #2686, opened 2015-12-16
Improve performance of d3_selectionPrototype.attr

vendethiel:
Current engines (tested on Firefox and Chrome) aren't *that* smart about `arguments` yet.
It is slow to assign to one while using the others, because they are bound: changing `name` also changes `arguments[0]`, which hinders the engines' ability to optimize the code.

The second issue is using a non-local value (and a `arguments` one at that) as the topic of a `for`-`in`, which is suboptimal.
In Chrome's profil, it warns with: "Not optimized: ForInStatement with non-local each variable".

### DIFF
--- a/src/selection/attr.js
+++ b/src/selection/attr.js
@@ -7,16 +7,16 @@ d3_selectionPrototype.attr = function(name, value) {
     // For attr(string), return the attribute value for the first node.
     if (typeof name === "string") {
       var node = this.node();
-      name = d3.ns.qualify(name);
-      return name.local
-          ? node.getAttributeNS(name.space, name.local)
-          : node.getAttribute(name);
+      var qualifiedName = d3.ns.qualify(name);
+      return qualifiedName.local
+          ? node.getAttributeNS(qualifiedName.space, qualifiedName.local)
+          : node.getAttribute(qualifiedName);
     }
 
     // For attr(object), the object specifies the names and values of the
     // attributes to set or remove. The values may be functions that are
     // evaluated for each element.
-    for (value in name) this.each(d3_selection_attr(value, name[value]));
+    for (var val in name) this.each(d3_selection_attr(val, name[val]));
     return this;
   }
 


mbostock:
Thanks for the suggestion. I suspect that this is not a noticeable performance improvement, but it’s probably worth adopting this type of change D3-wide. I’m not sure I’m willing to find and fix all of these places in D3 3.x, but I’d be willing to survey the new [d3 modules](https://github.com/d3) for 4.0 and adopt this optimization.

vendethiel:
It appeared in my profiler output, that's why I changed it in the first place. We call setAttribute (through attr) **very** extensively, it seems...

mbostock:
Yep, anything that touches the DOM is typically expensive; it should be much more expensive than re-assigning to an argument.

vendethiel:
Right, but at least it silences the chrome profiler warnings. The second thing I'm the most "sad" about is the array subclassing using __proto__, but I don't really have the time to "fork" and change everything for our codebase... Oh well, guess I'm stuck with this in the meantime.

Do you want me to close this PR, or are you interesting to take it on this repo?

mbostock:
The array subclassing is going away in 4.0; it’s not a backwards-compatible change so it has to wait until the next major release. See the [d3-selection repo](https://github.com/d3/d3-selection).

I don’t plan on merging this into 3.x, but I was going to keep this issue around as a reminder to go through the new [4.0 repositories](https://github.com/d3) and look for similar problems.

vendethiel:
Good news, thanks :)

mbostock:
The good news is that D3 4.0 is written in ES2015 modules and strict mode, thus this is no longer an optimization killer.

vendethiel:
Is this confirmed not to trigger perf. warnings in chrome?

mbostock:
See https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#3-managing-arguments

mbostock:
Also, d3-selection generally avoids re-assigning to arguments. See the new [selection.attr implementation](https://github.com/d3/d3-selection/blob/master/src/selection/attr.js), for example.

PR #2693, opened 2015-12-23
added uk_UA locale

ieaglle:
added uk_Ua locale as well as some tests


drimacus182:
@ieaglle Why isn't it merged into mbostock:master?

ieaglle:
@drimacus182 I've found a spelling error, canceled, and forgot about that. I probably should proceed with this.

drimacus182:
@ieaglle It would be nice)

mbostock:
I have added these to d3-format and d3-time-format and they will be published to npmcdn.com on the next patch releases.

PR #2695, opened 2015-12-29
Platlets

lokesh005:
Platlets is the extention of Pie-Chart on d3. Its a new chart having
attractive shape. If you want to know what the pie chart looks like
visit:http://bl.ocks.org/mbostock/3887235

For platelets visit: http://bl.ocks.org/lokesh005/1b23c84b68f5be134ff0

### DIFF
--- a/index8.html
+++ b/index8.html
@@ -1,0 +1,114 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>No. Of Victims</title>
+	<script src="http://labratrevenge.com/d3-tip/javascripts/d3.tip.v0.6.3.js"></script> 
+    <link rel="stylesheet" href="normalize.css">
+    <style>
+	  .tooltip{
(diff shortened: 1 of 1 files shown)


lokesh005:
what does it mean

mbostock:
Thank you for the contribution, but this would be better as an example (as you have already published on bl.ocks.org) rather than added to the git repository. Cheers!

lokesh005:
Hey but why not it should be in your galary as this svg is unique

lokesh005:
Please answer me

mbostock:
Examples are added to the gallery by anyone editing the wiki: it’s publicly editable. You don’t need to submit examples as pull requests.

That said, I’m not convinced this is an effective design for visualization. The large rounded corners subtract more from the larger arcs than the smaller arcs. I don’t see why this is better than a standard pie or donut chart.

lokesh005:
Okay thanks

PR #2708, opened 2016-01-11
Fix browserify key in package.json

ferlores:
if we want to define a different entry point for browserify, we should be using the browser key: https://github.com/substack/node-browserify#browser-field

The browserify key in package.json should be an object that is used to define globalt transforms: 
https://github.com/substack/node-browserify#browserifytransform

This fix will allow d3 to be used through cartero (https://github.com/rotundasoftware/cartero) which relies on browserify for creating bundles

Thanks!
F

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "url": "https://github.com/mbostock/d3.git"
   },
   "main": "d3.js",
-  "browserify": "d3.js",
+  "browser": "d3.js",
   "jspm": {
     "main": "d3",
     "shim": {


jfrazier:
+1

mbostock:
Fascinating. I wonder if this has been broken since #629 in May 2012.

ferlores:
Not sure, maybe it worked fine with an older version of browserify. 
Please let me know when you are publishing a new version with the changes. Thanks!

jfrazier:
@mbostock, do you know when this version will be published to npm? Needed for Browserify to work properly. Thanks!

mbostock:
Published 3.5.13.

jfrazier:
Awesome, thanks Mike!

PR #2714, opened 2016-01-20
Capitalize lines 100-103

jshen212:
Capitalized lines 100-103 to match other single-line comments


mbostock:
I’ll probably be rewriting this code for D3 4.0, so it doesn’t feel worth it to me to apply this change, sorry. Also, those comments are sentence fragments rather than sentences; complete sentences should be capitalized and have punctuation, but sentence fragments or single words are often done lowercase and without punctuation.

Issue #2665, opened 2015-12-03
When used as AMD still exports global


danez:
Why is d3 still adding the global d3 if used as AMD module?
https://github.com/mbostock/d3/blob/master/src/end.js#L1
In our case this causes a lot of troubles, as we are in the process of migrating a legacy system (which uses no module loader) to a new webpack base solution (with modules).
As both ways seem to write the global variable, d3 gets confused and lots of exceptions happen.

All other packages we use don't do that and properly use `define()` only.

mbostock:
Duplicate of #2645. See #1921 and #1693. To summarize: this was done because many D3 plugins assumed the existence of a global `d3`, so when AMD support was added and removed the global, those plugins broke.

The `d3` global will go away in 4.0 when everyone switches to the [D3 module pattern](http://bost.ocks.org/mike/d3-plugin/). This not a backwards-compatible change and thus requires being tied to the next major version.

danez:
Ahh thanks. Until version 4 gets released, I solved it by telling webpack not set `define` for d3.

```json
loaders: [{
    test: /d3\.js$/,
    loader: "imports?define=>false"
}]
```

Issue #2645, opened 2015-11-18
window.d3 despite being loaded as an amd module?


mindspank:
```javascript
if (typeof define === "function" && define.amd) this.d3 = d3, define(d3)
```

We are using requirejs to load d3, however it seems that d3 has started to define itself in global scope now. Is there a reason I'm missing here?

We usually have multiple third party plugins running our software and many of them are using d3.
With 3.4.X of D3 we never noticed any issue but now we are getting clashes between d3 versions in global scope.

mbostock:
Yes. See #1921.

mbostock:
This has been the case since [3.4.10](https://github.com/mbostock/d3/releases/tag/v3.4.10).

mindspank:
Ah ye, it was an odd edge case where we found it.

Module A loads D3js and shims in a sankey plugin.
Module B loads D3js and in the process overwrites the sankey plugin.
Module A tries to fire methods on the sankey plugin which has now been purged from window.d3

Cheers for the fast reply!

Issue #1693, opened 2014-01-10
Something wrong with the http://d3js.org/d3.v3.min.js package


kevinsigwart:
All of sudden all my applications that reference  http://d3js.org/d3.v3.min.js broke.  I'm getting a d3 reference not found error.  When I go to the downloaded version, that I downloaded a month or two ago, everything works fine.  Was there some new push that came out sometime this past week or two?

Thanks!

JohnGilmore100:
Is there a way to connect to 

HTTPS:://d3js.org/d3.v3.min.js  

as opposed to HTTP://d3js.org/d3.v3.min.js ?

JohnGilmore100:
Connecting to: HTTP://d3js.org/d3.v3.min.js  is OK

Where can I find a link to HTTPS:://d3js.org/d3.v3.min.js ?

cliveb:
Can't remember how I solved or hacked a d3 solution
http://navigator.xclimatechange.com/
Answer may be in the code https://github.com/cliveb/Gigaimpacts

Browser Vendors enforcing https over http notching up security for web
components keeps us on our toes.



--
clive || http://about.me/cliveboulton

cliveb:
I asked a Googler on Chrome Polymer/Web team how to solve this.

Polymer/Web Components don't have anything to do with security. Imports
behave no differently then trying to pull in JS from a different domain. If
you're getting errors in your console about linking to assets on different
domains then it may be because they haven't enabled CORS headers. The
easiest solution would be to serve the library from your own domain.

As for creating a custom element that uses d3, that should be fine.

--
clive || http://about.me/cliveboulton

mbostock:
Unfortunately, I haven’t yet setup HTTPS when using custom domains over GitHub Pages—I need to signup for Cloudflare or something similar. In the meantime, I recommend using a CDN that supports HTTPS, such as CDNJS:

https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.5/d3.min.js

JohnGilmore100:
Thanks all.

Mike - same feedback: 
* cloudflare: the latency seems to be OK, downloads reliable.
* I would like to suggest to you to also consider deploying d3 in https://ajax.googleapis.com

DatamineAnalytics:
This may or may not be related, but when developing on the file system in firefox (i.e. not using http-server or the like) the following (as recommended by https://d3js.org/) does not work for me:

[...]

PR #2715, opened 2016-01-20
Avoid global leak of d3 when AMD is used

stream7:
I'm using webpack to build a third party library and want to avoid exposing my lib's d3 to the global namespace since the client might already be using another version of d3 which I don't want to be overwritten.

Webpack supports both AMD and CommonJS but it will use [the first one](https://github.com/webpack/webpack/issues/883) so it ends up using the AMD definition here which has this extra "export" to the window object.

I think this exposure is not needed when d3 is used with AMD and it's trivial to be added in user code.

Thoughts?

### DIFF
--- a/src/end.js
+++ b/src/end.js
@@ -1,4 +1,4 @@
-  if (typeof define === "function" && define.amd) this.d3 = d3, define(d3);
+  if (typeof define === "function" && define.amd) define(d3);
   else if (typeof module === "object" && module.exports) module.exports = d3;
   else this.d3 = d3;
 }();


mbostock:
See #2665, #2645, #1921 and #1693. To summarize: this was done because many D3 plugins assumed the existence of a global `d3`, so when AMD support was added and removed the global, those plugins broke.

The `d3` global will go away in 4.0 when everyone switches to the [D3 module pattern](http://bost.ocks.org/mike/d3-plugin/). This not a backwards-compatible change and thus requires being tied to the next major version.

PR #2723, opened 2016-01-27
selection.style: Allow map generating functions as a single argument

sztanko:
I would like to call selection.style(), with a single argument that is a map generating function. Something like this:
```
selection.style(function(d) {return {'fill': someScale(d.x), 'stroke': someOtherScale(d.y)};});
```

This will make it usable with functions that generate [inline styles](https://facebook.github.io/react/tips/inline-styles.html) for React.js.

### DIFF
--- a/src/selection/style.js
+++ b/src/selection/style.js
@@ -10,7 +10,13 @@ d3_selectionPrototype.style = function(name, value, priority) {
     // functions that are evaluated for each element. The optional string
     // specifies the priority.
     if (typeof name !== "string") {
-      if (n < 2) value = "";
+      if (n < 2) {
+        value = "";
+        if (typeof name === "function") {
+          this.each(d3_selection_function_style(name));
+          return this;
+        }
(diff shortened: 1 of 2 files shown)


mbostock:
See [d3-selection-multi](https://github.com/d3/d3-selection-multi) for 4.0, in particular [*selection*.styles](https://github.com/d3/d3-selection-multi/blob/master/src/selection_styles.js). It’s still a work-in-progress and I haven’t written the documentation yet, but if you’d like help, please send a pull request over there.

The new API is almost exactly what you describe, though it has a distinct name to disambiguate from the single-value case. For example:

```js
selection.styles({"fill": "red", "stroke": "blue"});
selection.styles(function() { return {"fill": "red", "stroke": "blue"}; });
```

I’m probably not going to include d3-selection-multi in the default build of 4.0, for the sake of parsimony and keeping things small, but it should be available if people want to add it in.

PR #2730, opened 2016-02-06
Section for non-technical readers

srsjake:
To make D3 more approachable to a larger number of users—from informed developers to curious passersby and new and interested visitors—I have added a new section to the README. The section is entitled What to Use D3 for and When to Use It. As the name suggests, the new section, a brief paragraph, explains in plain and straightforward language what D3 is used for and when one should consider using it in a project. 

Other repositories may begin to use this new informative and catchy section soon, as my colleagues contact open-source technologies in an effort to make them more accessible to the tens of thousands of new users learning to code and thousands joining GitHub every day.

Thank you!


mbostock:
I appreciate the suggestion, but I think this longer material is better suited to [the publicly-editable wiki](https://github.com/mbostock/d3/wiki) rather than the README. Thanks.

PR #2732, opened 2016-02-08
Non-developer README contribution

Jing-Ran:
I have written a clear, easy-to-comprehend introduction for d3.js. I believe this introduction, free of complex technical terms, will swiftly and thoroughly inform anyone about the use-cases, merit, and advantages, including the celebrated features, of d3.js. Seasoned developers, new programmers, and even nontechnical visitors will benefit from the clear language, exposing d3.js to a broader range of users.


mbostock:
Thanks for the suggestion; however, I prefer the current text.

PR #2744, opened 2016-02-20
Variable update

tvictory:
Updated variables in the color class to add clarity to the variable names at a quick glance, as the subject matter can be confusing as it is. The appropriate documentation will be updated if this pull request is accepted.

### DIFF
--- a/d3.js
+++ b/d3.js
@@ -1580,84 +1580,84 @@
     return this.rgb() + "";
   };
   d3.hsl = d3_hsl;
-  function d3_hsl(h, s, l) {
-    return this instanceof d3_hsl ? void (this.h = +h, this.s = +s, this.l = +l) : arguments.length < 2 ? h instanceof d3_hsl ? new d3_hsl(h.h, h.s, h.l) : d3_rgb_parse("" + h, d3_rgb_hsl, d3_hsl) : new d3_hsl(h, s, l);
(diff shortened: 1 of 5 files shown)


alanvanbuuren:
Thnx for the new variable name's

mbostock:
This code has already been rewritten for D3 4.0 (see [d3-color](https://github.com/d3/d3-color)) so I will not be accepting style-only changes.

mbostock:
Apologies for the terse response; I was distracted by my three-year-old. I appreciate your offer to make the code clearer and I’m sorry I’m not able to accept your request.

Code clarity is a matter of taste, and I don’t feel that these changes are universal improvements; I prefer shorter variable names when the meaning is already apparent from context, such as the R, G and B channels of the RGB color space.

That said, I know that sometimes I try too hard to make the code concise and that can make it harder to follow, and there could be some improvements I would accept. I’ve implemented a few changes in d3-color for 4.0 to make it more readable already, such as separating color space conversion, as in `hslConvert(color)`, from the color specification, as in `hsl(h, s, l, a)`.

If there are other changes you think might improve readability, please let me know. Although it may be best to smart small and discuss it early. Thank you.

tvictory:
I appreciate your feedback and will consider taking a look at things in d3-color 4.0 instead then.

PR #2746, opened 2016-02-22
minor performance improvement to transition#attr

timruffles:
Chrome can't optimise the `for ... in` loop here as `value` is a non-local variable (i.e has been passed in from outside the function).

Below you can see before it was un-optimisable, after it optimises ok (ignore timings, they're not comparing like with like).

### Before
before
### After
after

### DIFF
--- a/src/transition/attr.js
+++ b/src/transition/attr.js
@@ -10,7 +10,8 @@ d3_transitionPrototype.attr = function(nameNS, value) {
     // For attr(object), the object specifies the names and values of the
     // attributes to transition. The values may be functions that are
     // evaluated for each element.
-    for (value in nameNS) this.attr(value, nameNS[value]);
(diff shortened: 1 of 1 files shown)


mbostock:
This code has been completely rewritten for D3 4.0; see [d3-transition](https://github.com/d3/d3-transition). I thank you for the request, but I’m not willing to accept such small changes for 3.x.

In 4.0, all the code is strict mode, which eliminates the optimization killer of reassigning to arguments in functions that also reference arguments.length. Also, the multi-value map methods have been extracted into separate functions to reduce ambiguity; see [d3-selection-multi](https://github.com/d3/d3-selection-multi). I haven’t written a d3-transition-multi yet, though.

timruffles:
Ok, makes total sense! :)

PR #2747, opened 2016-02-24
Fire "dragend" on "touchcancel"

kristofsch:
The dragend is not properly triggered when using a touch a "touchcancel" event was triggered.

This caused our application to be unable to scroll in some cases when implementing scrolling behaviour inside a graph.
On iOS this can be reproduced by using the 4 or 5 finger swipe on the application.


mbostock:
The new [d3-drag behavior](https://github.com/d3/d3-drag) for 4.0 now handles the touchcancel event.

PR #2759, opened 2016-03-07
Better ticks for subsecond domains.

### LINKED ISSUE #1757: d3 time scale fails to rescale correctly at millisecond intervals
See my stackoverflow post here:

http://stackoverflow.com/questions/22051638/d3-zooming-panning-behavior-changes-at-millisecond-intervals

After further investigation, I discovered the problem is in the time scale, not the zoom behavior. See my edit.

shanths:
Rather than starting the ticks on the minimum domain value, round up based on
the step size. Fixes #1757.


mbostock:
This looks fixed in 4.0.

https://tonicdev.com/56c79a2c9b27fc0c00ea757a/5787e92e6ebb1c140083a7d9

PR #2782, opened 2016-03-31
function as single argument for selection.attr

sunrize531:
This could be a big time saver and optimisation, since you won't need to generate/call separate functions for each attribute. For lazy developers who don't use predefined functions and for anyone else.

P.S. Unfortunately I couldn't install dependencies on my machine, and this might be a reason that tests don't pass as well, and I can't build. Which version of node should be used?

### DIFF
--- a/src/selection/attr.js
+++ b/src/selection/attr.js
@@ -12,6 +12,12 @@ d3_selectionPrototype.attr = function(name, value) {
           ? node.getAttributeNS(name.space, name.local)
           : node.getAttribute(name);
     }
+    
+    if (typeof name === "function") {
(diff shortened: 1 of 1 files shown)


mbostock:
This code has been completely rewritten for D3 4.0 in [d3-selection](https://github.com/d3/d3-selection). The multi-value map methods have been extracted into a separate repository, [d3-selection-multi](https://github.com/d3/d3-selection-multi), which will be an optional plugin. The new *selection*.attrs (the new multi-value map method) supports taking a function.

(Also, a change like this have a huge ripple effect, since there are many similar methods (such as *selection*.style and *transition*.attr) that should be kept consistent.)

Issue #2025, opened 2014-09-18
d3.geo.path doesn't handle zero-area polygons correctly


stvno:
d3.geo.path sometimes renders polygons where all points are in a straight line (so no *inside*) as an "infinite" area with a line shaped "hole". The polygons are created by TileStache from a postgis database and served as topojson. The polygons start out as valid RHR geometries and are turned into lines due to the shift from floating point space to integer space; see the drawing below.
![example](https://cloud.githubusercontent.com/assets/394323/4317788/2b755bb8-3f17-11e4-9bb7-8c7e235cba17.png)

If the order of the last three points in the topojson is changed the feature is drawn correctly. I've setup a page showing the difference here: http://research.geodan.nl/sites/vectortiling/test/faultytile.html

Is there a correct clockwise order for points on a line? If so it might be that TileStache is wrong here serving the points in the wrong order, but following the above drawing the points are in a valid order. I would think that if a polygon has an area of 0, the order of points doesn't matter and D3 shouldn't try to create an infinite area with a zero area hole in it.

mbostock:
Fixed in 3.5.17; see #2784 by @lukasappelhans.

Issue #2024, opened 2014-09-17
Caspian Sea winding order bug


mbostock:
Discovered this with the new graphics pipeline, but I suspect that it would occur with master as well:

[![screen shot 2014-09-17 at 8 59 40 am](https://cloud.githubusercontent.com/assets/230541/4306770/aa4ffb9e-3e83-11e4-9583-e5a8cc4cde9d.png)](http://bl.ocks.org/mbostock/29dcd63773f78f7bad26/b4bb2326a27ad489731143bd4dfb4f690ae079ea
)

The pipeline is extremely short (yay!):

```js
var render = d3.geo.pipeline()
    .source(d3.geo.jsonSource)
    .pipe(d3.geo.rotate, 0, 2.85, 1.42)
    .pipe(d3.geo.clipAntimeridian)
    .sink(d3.geom.contextSink, context);
```

mbostock:
Related #2025. Here’s the 3.x version:

http://bl.ocks.org/mbostock/e114d89ab8440cfc8539ce2414eceb95

Sure enough, it looks like the world.objects.land feature is busted. If you look just at the first polygon, which contains Russia:

![screen shot 2016-05-04 at 1 10 50 pm](https://cloud.githubusercontent.com/assets/230541/15027639/a9370978-11f9-11e6-90af-db3f955a9969.png)

You’ll see that the coordinates look suspicious. The rings of the polygon have the following lengths:

1. 4
2. 4
3. 1322
4. 52

Clearly the third polygon with the most coordinates should be the exterior ring. Furthermore, the area of these polygons:

1. 2.2953421996046487e-7
2. 12.566370560342287
3. 1.9770847424837752
4. 12.556828645500577

So, the first polygon does appear to be clockwise, and perhaps that’s why it’s being chosen over the third. 

The first polygon appears to correspond to [Lake Tana](https://en.wikipedia.org/wiki/Lake_Tana) and at this rotation it ends up spanning the antimeridian, hence the winding order calculation being off.

Switching the first and third polygon fixes the problem, although that assumably causes Lake Tana to not be filled (which is not noticeably because I believe it’s a degenerate line).

Another way to fix it is to swap the two middle coordinates of Lake Tana, which currently looks like this:

```
33.96453964539646,9.464925113251127
33.97533975339755,8.685250682806824
33.96093960939609,9.583004982449822
33.96453964539646,9.464925113251127
```

mbostock:
[...]

PR #2784, opened 2016-03-31
Possible fix for #2025, handle zero-area polygons better

lukasappelhans:
Hi!

I was hit by what seems to be #2025 when developing a couple of canvas-based maps. This change fixes it for me.

I have no idea if this is correct or makes any sense in general, but I have not found any regressions so far.

Thank you,

Lukas

### DIFF
--- a/src/geo/point-in-polygon.js
+++ b/src/geo/point-in-polygon.js
@@ -67,5 +67,5 @@ function d3_geo_pointInPolygon(point, polygon) {
   // from the point to the South pole.  If it is zero, then the point is the
   // same side as the South pole.
 
-  return (polarAngle < -ε || polarAngle < ε && d3_geo_areaRingSum < 0) ^ (winding & 1);
+  return (polarAngle < -ε || polarAngle < ε && d3_geo_areaRingSum < -ε) ^ (winding & 1);
 }


martgnz:
@mbostock please can you take a look at this? We are going to release a Canvas-based map library and  don't want to rely on forks.

mbostock:
The math here is extremely nuanced and reviewing this change, albeit a tiny one, is deceptively difficult. I’d love @jasondavies input here, but I haven’t heard from him recently. I’d like to review this change but I have a lot of work already on my plate trying to get D3 4.0 released ASAP, so I’m not sure when I’ll be able to dive into it.

mbostock:
Can you share the data that led to the winding-order problem? That would help me test this change.

I just fixed a related bug #2024 which turned out to be in topojson.merge: it was selecting the wrong exterior arc after merging the polygons. That arc happened to be degenerate, with the wrong winding order. I’ve fixed topojson in [1.6.26](https://github.com/mbostock/topojson/releases/tag/v1.6.26) and I’ve pushed new data files to [bl.ock 4090846](http://bl.ocks.org/mbostock/4090846) which is used by many of my D3 examples.

lukasappelhans:
Okay, unfortunately this doesn't seem to be the issue here. I updated this block to use the latest official topojson and d3 releases. You can see that the bounds are fine now, but at least one polygon "leaks".

http://bl.ocks.org/lukasappelhans/967f744f80ff2209ad946570063210f4

mbostock:
Can you also post how you made municipios.json?

mbostock:
Strangely, I’m not able to reproduce the bounding box or rendering problem with the data you provided. It seems to work fine?

[screen shot 2016-05-04 at 3 45 49 pm](http://bl.ocks.org/mbostock/d5deee4c249f101a17f1dead0089109e/b62abf399225d0504118fa9bc11a2c9d71079440)

The bounding box returned by d3.geo.bounds is:

```json
[
  [-18.161305014722235, 27.637839000751285],
  [4.3277847236038625, 43.79237956913792]
]
```

This seems reasonable. It looks like when rendered it’s not exactly capturing all of the Canary Islands, so I suppose there could be a small bug in d3.geo.bounds, but it looks approximately correct.

lukasappelhans:
Yes, d3.geo.bounds is fine with the other PR merged. The rendering also works as long as I keep dynamic simplification disabled. Once I turn it on, it paints everything in one color.

martgnz:
BTW, I just created a new TopoJSON using the latest version and still "leaks": http://bl.ocks.org/martgnz/7889428fcad5eb8b43eec7c71ddb90af

I made this json with:
```bash
topojson municipios.shp -o municipios.json
```
You can download the original shapefile here: https://drive.google.com/file/d/0B_UyUs3UduSUNkZiMmdJRlVJTlU/view?usp=sharing

mbostock:
Okay, I think I’ve figured out the discrepancy here: when reproducing your example, I removed the call to topojson.presimplify, so it seems like that’s still introducing artifacts.

mbostock:
Well, more precisely, it isn’t that topojson.presimplify is introducing artifacts, it’s that this enables dynamic simplification in the client using your simplify transform (which is presumably based on my [dynamic simplification example](https://bl.ocks.org/mbostock/6245977)):

```js
d3.geo.transform({
  point: function(x, y, z) {
    if (!z || z >= settings.area) {
      this.stream.point(x, y)
    }
  }
})
```

So I assume what’s happening is that the dynamic simplification is introducing degenerate polygons, some of which are interpreted having the wrong winding order.

mbostock:
The other thing is that dynamic simplification is really designed to operate on *projected* (planar) geometry, not spherical geometry. The idea is that you project the non-simplified geometry down to the plane, and then filter the planar geometry to the desired simplification. Your code does it in the opposite order: it filters the spherical geometry and then projects down to the plane. My guess is that’s more likely to introduce degenerate polygons like this (and it’s also a lot more expensive).

mbostock:
So yeah, this is definitely the same as #2025. This is an example of a dynamically simplified polygon:

```
-13.487800541155952 29.40322484792305
-13.487125861717008 29.40322484792305
-13.486900968570694 29.40322484792305
```

The latitudes are identical, so this has obviously collapsed to a straight line.

mbostock:
Okay, I’ve tested your change and it looks good! Will release shortly.

lukasappelhans:
Awesome, thank you!

Yes, I agree that projecting the topojson beforehand would be better and faster. However, there is (correct me if I'm wrong) no way to use custom projections (e.g. rveciana/d3-composite-projections) when using the cli client. The other problem I see is responsiveness, since we use different projection parameters for mobile vs desktop. Other than that, I should at least add support for projected topojsons.

I wonder why the simplification produces a straight polygon though. Shouldn't this be avoided in any case?

mbostock:
I recommend loading different pre-projected data if you need it. Projection
is especially slow on mobile devices—I can’t imagine a 12MB unprotected
dataset displaying, although maybe the latest devices are fast enough these
days.

Another consideration is that presimplify assumes planar geometry by
default: it uses planar triangles for Visvalingam simplification. If you
give it spherical geometry the math is wrong. You need to pass the
spherical triangle area function as the second argument rather than using
the default planar area. (This happens automatically on the command line.)
The difference may not be huge depending on the size of your dataset
relative to the size of the Earth though, but it just exacerbates the
problem of simplifying before projecting.

mbostock:
I’m pretty sure you can use custom projections from the command-line, but
you need to require them (and npm install them) first. The projection
definition is just JavaScript that gets evaluated and returns the
projection object, so anything that can be run in Node should be available
for you to use. You can even define the projection definition in a separate
JavaScript file and require that if escaping it through the command-line is
a pain.

That said this needs better documentation and the design of the
command-line tools needs work (specifically decoupling into smaller UNIX-y
tools). It’s on my long list of things to do after I finish D3 4.0.

PR #2786, opened 2016-04-04
Added Czech locale

zdenekhynek:


mbostock:
These have been added to d3-format and d3-time-format.

PR #2795, opened 2016-04-13
Improve regex logic in d3_time

isaacl:
This speeds up d3_time_parse, a hot function.


isaacl:
What do I need to do to land this?  It's a small, self-contained change that improves performance in c3 charts (which use d3)

mbostock:
Two things.

One is a benchmark demonstrating the degree of performance improvement. I normally suggest JSPerf but that appears to be down due to spam. So something that runs in the browser would be sufficient.

Two is that D3 3.x is basically frozen at this point as I am working on releasing D3 4.0 in the next few months. So this pull request should really be re-targeted at the [d3-time-format repository](https://github.com/d3/d3-time-format).

mbostock:
I wrote a benchmark for this over in [d3-time-format](https://github.com/d3/d3-time-format)’s [match-not-exec branch](https://github.com/d3/d3-time-format/tree/match-not-exec). However, I don’t see any impact on performance. If you still see a performance improvement using *string*.match instead of *regex*.exec, please let me know how to reproduce that gain and I will consider merging the match-not-exec branch (or implementing other optimizations!). Thanks.

PR #2798, opened 2016-04-15
More comprehensive workaround for IE9.

maliang03:


mbostock:
D3 3.x is basically frozen at this point as I am focused on releasing D3 4.0 in the next few months. Also, D3 is not a compatibility layer, and I am increasingly inclined to remove patches like this and focus my efforts on modern browsers. So a fix like this is probably best implemented as a shim, either for D3 or for IE9 more generally.

PR #2803, opened 2016-04-20
Bug in safari if use as define module

gunins:
If in safari use as amd, after compiler.inside scope `this` is `undefined`, also looks like same errors apears on IE as well. I added `call(window||this)` at the end, and this fix working fine.

### DIFF
--- a/src/end.js
+++ b/src/end.js
@@ -1,4 +1,4 @@
   if (typeof define === "function" && define.amd) this.d3 = d3, define(d3);
   else if (typeof module === "object" && module.exports) module.exports = d3;
   else this.d3 = d3;
-}();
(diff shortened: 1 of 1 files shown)


mbostock:
The only case where `this` is undefined is if you enable strict mode. Are you building your own d3.js file or concatenating multiple files together? At any rate, D3 3.x is basically frozen as I am working on releasing D3 4.0 in the next few months. D3 4.0 is implemented using ES2015 modules and uses [Rollup](https://rollupjs.org) to produce a UMD bundle.

gunins:
Yes it is in strict mode and bundled by r.js. Will Look forward on version 4.

PR #2804, opened 2016-04-20
chinese locale

vanshady:
Changed the Chinese locale to make it closer to the Chinese custom and follow the [localeplanet rules](http://www.localeplanet.com/java/zh/). 
Added its corresponding test.


tianxuzhang:
nice +1

mbostock:
Thanks. I’ll port this to d3-time-format for 4.0.

PR #2812, opened 2016-04-26
Add d3.create and selection.adopt.

ginogabriel:


mbostock:
D3 3.x is basically frozen as I work on [D3 4.0](https://github.com/mbostock/d3/tree/4) which will be released in the next few months. If you want to propose changes to the selections API, please file an issue over in [d3-selection](https://github.com/d3/d3-selection). However, I’m extremely conservative about adding new API methods, and I’m not inclined to add these methods as-is, especially without discussion. Thanks!

PR #2984, opened 2016-09-27
Native map

Spencer8989:


mbostock:
Not sure if this was an intentional PR, but it applies to 3.x code which was moved in 4.0. Also due to a desire to remain compatible with older platforms that do not yet support native collections, I decided not to use include these commits in the 4.0 release.

PR #3019, opened 2016-11-04
fixed wye decription

askmike:


mbostock:
Do you mind changing it to “a Y shape” instead of “an Y-shape symbol”? I would, but I am on mobile.

askmike:
done. I got the text from the [d3-shape README](https://github.com/d3/d3-shape/blob/master/README.md#symbolWye). Want me to change that as well?

mbostock:
My thanks!

mbostock:
Nah, that README looks fine. The text here is abbreviated.

PR #3057, opened 2017-01-26
CHANGES.md --> link v3, v4 tickSizeInner & tickSizeOuter syntax

micahstubbs:
This PR adds the d3 v3 `axis.innerTickSize` and `axis.outerTickSize` so that a user searching *CHANGES.md* for the v3 syntax will find a mention of it and a pointer to the new v4 syntax

screen shot 2017-01-26 at 12 00 14 pm

here are the [d3 v3 Axes API docs](https://github.com/d3/d3-3.x-api-reference/blob/master/API-Reference.md#axes), for reference


mbostock:
Merged. Thanks.

PR #3063, opened 2017-02-03
Attempt at zoom.translateExtent.

jraah55:
Doesn’t work correctly because the translateExtent may need to be dependent on
scale, or the zoom behavior may need to know the visible area. Related #1084.


mbostock:
With apologies but I am not accepting pull requests for D3 3.x.

yach7:
#3062

PR #3190, opened 2018-01-18
Update desc of color.toString to match v4 functionality

zanarmstrong:
Super tiny change in documentation. 

Looks like the functionality of color.toString changed between v3 and v4. v3 returned hex representation and v4 returns string according to CSS spec. But, this description got missed in the md file. 

I wasn't sure if you'd prefer "returns string representation of rbg values" or "returns string according to CSS spec" or something else. Mostly just wanted to flag the need for this tweak in the documentation. 

thanks!


mbostock:
Great, thank you!

PR #3283, opened 2018-03-23
Format Promise in CHANGES.md to be consistent

oscarmorrison:


mbostock:
Thanks for the suggestion, but I prefer the original formatting.

oscarmorrison:
No worries @mbostock thanks for your consideration. 

I was mainly try to keep it consistent with this style

```
var x = d3.scaleBand()
    .domain(["a", "b", "c"])
    .range([0, width]);
```

PR #3299, opened 2018-06-21
Update License to cover 2018

josephlin55555:


Loksly:
#3288 #3259 and many more.

PR #3302, opened 2018-10-29
Environment

sixwolves:
sdd


curran:
@sixwolves This is a strange PR. What does it mean?

PR #3303, opened 2018-10-30
Include more metadata in bower.json; fixes #1707.

dear123dear:


bin/bower
```diff
@@ -1,10 +1,17 @@
 #!/usr/bin/env node
 
-var fs = require("fs");
+var fs = require("fs"),
+    pkg = require("../package.json");
```
Loksly: instead of require you may use `pkg = JSON.parse(path.resolve(__dirname, '..', 'package.json')); ` or something like that. Require not only parses but executes

PR #3318, opened 2018-12-05
NPM command was not visible enough , so quoted it.

arnab-shuvo:
NPM command was not visible enough to user. Infact at first I missed it. So tried to make it more visible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,13 @@
 
 ## Installing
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,7 +14,13 @@
 
 ## Installing
 
-If you use npm, `npm install d3`. Otherwise, download the [latest release](https://github.com/d3/d3/releases/latest). The released bundle supports anonymous AMD, CommonJS, and vanilla environments. You can load directly from [d3js.org](https://d3js.org), [CDNJS](https://cdnjs.com/libraries/d3), or [unpkg](https://unpkg.com/d3/). For example:
+If you use npm, Use simple npm command. For Example:
+
+```html
```
stevemao: should be shell

mbostock:
Thank you for the suggestion. I prefer the current wording.

Issue #3131, opened 2017-08-08
Add sideEffects: false to package.json. 


TheLarkInn:
Hello D3 Family!!!!! @TheLarkInn from [webpack](https://github.com/webpack) here! 

We have been working on a new feature for webpack 4+ which allow's library authors to add a special field to package.json allowing them to declare that their libraries that have "reexports" in them do not contain side effects!

#### Tiny Backstory
Because webpack implements the Harmony Module Specification to spec, we include code that was reexported (even if it wasn't used). This is because all code [per spec] must be eval'd). The beautiful news is that we do not have to implement _exactly to spec_. So we have created what is called "pure-module". webpack as imports are being resolved will also check to see if the module is pure (which is indicated through the `pure-module` flag in the package.json fields). Therefore, there would be no bundle size difference between the following examples:

```
import {bottomAxis} from "d3-axis";
```

and 

```
import bottomAxis from "d3-axis/direct/path/to/d3AxisBottom"
```

#### Proposal
If there are no side-effects within the D3 packages in places where re-exports occur, then I'd love to propose that all D3 packages add `pure-module: true` to their package.json. 

#### Pros
Developers (using webpack) get huge perf wins by doing nothing.

#### Cons
Bit more work to get it all done. 

Please let me know if you have any questions, the webpack team is really anxious to help in any way if we can (that means, submitting PR's, adding issues to individual repo's to track, etc). 

sideEffects: true
- [x] [d3-format](https://github.com/d3/d3-format) ([src/defaultLocale.js](https://github.com/d3/d3-format/blob/master/src/defaultLocale.js))
- [x] [d3-time-format](https://github.com/d3/d3-time-format) ([src/defaultLocale.js](https://github.com/d3/d3-time-format/blob/master/src/defaultLocale.js))
- [x] [d3-transition](https://github.com/d3/d3-transition) ([src/selection/index.js](https://github.com/d3/d3-transition/blob/master/src/selection/index.js))

sideEffects: false
- [ ] [d3](https://github.com/d3/d3)
- [x] [d3-array](https://github.com/d3/d3-array)
- [x] [d3-axis](https://github.com/d3/d3-axis)
- [ ] [d3-brush](https://github.com/d3/d3-brush)
- [x] [d3-chord](https://github.com/d3/d3-chord)
- [x] [d3-color](https://github.com/d3/d3-color)
- [x] [d3-contour](https://github.com/d3/d3-contour)
[...]

PR #3319, opened 2018-12-19
improve docs on modules

stevemao:


stevemao:
@envision thanks!, I'd reword it to: **NOTE: This will bundle every module in D3**

To try to avoid mentioning tree shaking.

milospanasiuk:
Hello there,

Can someone guide on how to update the wiki with an additional plugin link? Tried various solutions but none was successful. 
When creating a pull request, the wiki is not detected from forked repo since its repo for itself.

curran:
I believe the wiki was locked as a result of abuse/spam. Take a look at the age of the entries there, they are all quite old now.

This may be the closest replacement for listing plugins/packages: https://github.com/wbkd/awesome-d3

mbostock:
Thank you for the suggestion. The behavior of tree-shaking is dependent on the bundler you are using; for example Rollup does tree-shaking fairly intelligently using static analysis. Also, tree-shaking may be affected by adoption of the sideEffects flag #3131 

While I know this PR is well-intentioned, I do not want to take responsibility for documenting how bundlers behave. People should refer to their preferred bundler’s documentation instead.

PR #3331, opened 2019-04-25
- Ignore `package-lock.json`

brettz9:


mbostock:
D3 modules will soon be adopting explicit files when publishing. #3335

PR #3332, opened 2019-04-25
- Provide prebuilt ESM distributions for out-of-the-box browser use

brettz9:
As with pre-built binaries, it is convenient to be able to use a good tool without build steps. With this PR, the ESM distribution can be used which, unlike the UMD distribution, does not require one to pollute one's HTML with extra, non-modular, script tag dependencies, and also ensures one's JavaScript is free of "magic" such that there are no globals and one can trace--starting with any file--where the variables it uses are defined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,13 @@ Or import everything into a namespace (here, `d3`):
 import * as d3 from "d3";
 ```
 
+You may also directly point to the bundled version if you are only concerned
+with supporting modern browsers:
+
+```js
+import * as d3 from "./node_modules/d3/dist/d3-esm.js";
+```
+
 In Node:
 
 ```js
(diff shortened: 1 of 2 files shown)


Ntropish:
This would be really helpful for me right now. D3 distributions try to use global `this` which is undefined and causes an error which makes them unloadable.

mbostock:
If you want to consume D3 as an ES module, you can use the `module` entry defined in the package.json, which is:

https://github.com/d3/d3/blob/master/index.js

I would like the D3 modules to ship minified ES module bundles for more efficient consumption in browsers, but the minified bundles should not inline transitive copies of their dependencies so that you can import D3 modules separately without duplicating code. If you’d like to help me get that working, please discuss over in #3336. Thank you!

PR #3339, opened 2019-08-12
fix: fix d3-scale-chromatic images URLs in CHANGES.md

severo:


severo:
I will change the strategy: better to link to the d3-scale code pinned at the moment of the 4.0 release

PR #3366, opened 2020-01-28
Support esm on node with conditional exports

sebamarynissen:
Since 13.6 node is able to load esm without the `--experimental-modules` flag, and since 13.7 node supports [conditional exports](https://nodejs.org/api/esm.html#esm_conditional_exports) unflagged as well.

While node can import cjs code from esm code just fine, currently only
```js
import d3 from 'd3';
```
is possible. If you try
```js
import { line } from 'd3';
```
you get
```
import { line } from 'd3';
         ^^^^
SyntaxError: The requested module 'd3' does not provide an export named 'line'
```
because Node doesn't support named exports from cjs modules.

I modified the build config to build an esm file for node as well and made sure node can find it using the conditional exports.

I am aware that the node version of d3 simply exports the individual `d3-*` modules whereas my build fully includes them in the code. I think that eventually the conditional exports should be used in each of the `d3-*` modules. I just wanted to add this as a PoC however to open up the discussion without going through all the work of implementing the conditional exports in each of the submodules.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,8 +20,8 @@
   "jsdelivr": "dist/d3.min.js",
   "module": "index.js",
   "exports": {
-    "require": "./dist/d3.node.js",
-    "import": "./dist/d3.node.mjs"
+    "import": "./dist/d3.node.mjs",
+    "default": "./dist/d3.node.js"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -19,6 +19,10 @@
   "unpkg": "dist/d3.min.js",
   "jsdelivr": "dist/d3.min.js",
   "module": "index.js",
+  "exports": {
+    "require": "./dist/d3.node.js",
+    "import": "./dist/d3.node.mjs"
+  },
   "repository": {
     "type": "git",
     "url": "https://github.com/d3/d3.git"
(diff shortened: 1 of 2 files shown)


mbostock:
Neat! The future is here. I’ll need to do some testing but this looks good at first glance.

sebamarynissen:
In the meantime it was discovered that we can silence the warning that conditional exports gives for `require`, see https://github.com/mrdoob/three.js/pull/18498#issuecomment-579962472 and https://github.com/chaijs/chai/pull/1317#discussion_r372633273. I'll push a new version.

kane2266:
D1

curran:
Curious what remains to be done here?

mbostock:
Do we want to add type: module here, too? And can we use the existing ES module entry point for environments that support it?

```
  "main": "./dist/d3.node.js",
  "exports": {
    "import": "./index.js",
    "require": "./dist/d3.node.js"
  },
  "type": "module"
```

sebamarynissen:
From memory, I think this would work *except* when you're using named imports on node 13.6 with the `--experimental-modules` flag since it doesn't support conditional exports, causing it to import the main field, which points to the cjs build that can't provide named exports. Since node 13 is EOL though I guess this can be ignored.

PR #3440, opened 2020-08-27
Add missing periods in v6 docs sentences

curran:
I couldn't help but notice the inconsistency that a few sentences were missing periods. This PR adds them. Very minor.

Really excited for v6!


mbostock:
The grammatical issue is more that we’re mixing sentences (which should have periods) and sentence fragments (which shouldn’t). I’ll take another pass at this section and see if I can make it better.

PR #3521, opened 2021-07-30
Update Dependencies 

mmorel-35:
Hi! I have used dependabot to update the current dependencies.


Fil:
thanks, but we'll follow our usual process for this

PR #3562, opened 2022-03-29
Updated API Reference Link in README.md

aedmondson:
The API Reference link at the top of README.md used "master" instead of "main". GitHub would redirect to the correct page, but with a warning message.


mbostock:
Thanks!

PR #3576, opened 2022-07-09
Repair package.json compatibility issues with react-native-cli

friyiajr:
The react native CLI requires the `./package.json` key be filled in for libraries it uses. Without this being filled in, we have no way of starting the Bundling Server. Several libraries are updating to compensate for this (See: https://github.com/react-native-community/cli/issues/1168).


mbostock:
This should be fixed in upstream tooling, not here. Sorry.

PR #3659, opened 2023-06-07
Add short "D3 in React" section

tophtucker:
I know you removed the TODO but I was already trying to fill it in! I think just making the distinction of modules that touch the DOM and those that don't was super clarifying for me personally when I figured that out. And I always forget the most basic ref pattern (and still might've messed it up here). I don't think we should get into updating or interactivity or whatever, but I think just this much goes a long way toward demystifying (and showing just the most basic best practices).

image

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -256,3 +256,49 @@ import {mean, median} from "d3-array";
 ```
 
 TypeScript declarations are available via DefinitelyTyped.
+
+## D3 in React
+
+Most D3 modules (like d3-scale, d3-array, d3-color, d3-format, and d3-random) don’t interact with the DOM, so there is no difference when using them in React. You can use them in JSX to do purely declarative visualization, like this dot plot of a normal distribution:
+
+```jsx
+import {scaleLinear, extent} from "d3";
+
+function DotPlot({data, width}) {
+  const x = scaleLinear(extent(data), [5, width - 5]).nice(true);
+  return (
+    <svg width={width} height="4">
+      {data.map((d, i) => (
+        <circle key={i} cx={x(d)} cy="2" r="2" />
+      ))}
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -256,3 +256,49 @@ import {mean, median} from "d3-array";
...
+
+function DotPlot({data, width}) {
+  const x = scaleLinear(extent(data), [5, width - 5]).nice(true);
+  const ref = useRef();
+  useEffect(() => {
+    if (!ref.current) return;
+    select(ref.current).append("g")
+        .attr("transform", "translate(0, 4)")
+        .call(axisBottom(x).ticks(5));
+  }, [x]);
```
mbostock, line 293: This will append a new axis every time it renders, right? Because every time DotPlot renders, it’ll create a new `x` scale, triggering this affect. I think either you have to remove the axis you added in a disposal hook for the useEffect, or you have to use a join to replace the previous axis (if any) when the effect runs. 🤔

mbostock, line 293: Maybe this would work? I think you can also drop the ref.current test because if ref.current is null, then select(null) will be empty and the code won’t do anything. (Maybe that’s too magical though…)

```suggestion
  useEffect(() => {
    const g = select(ref.current).append("g")
        .attr("transform", "translate(0, 4)")
        .call(axisBottom(x).ticks(5));
    return () => g.remove();
  }, [x]);
```

mbostock: This worked for me…

https://codesandbox.io/s/objective-mendeleev-6oj5kg?file=/src/DotPlot.jsx:0-606

tophtucker: oops yeah. great example of why i personally need this section

mbostock (approved):

PR #3666, opened 2023-06-09
two more anchors… hopefully the last

Fil:


docs/d3-geo/projection.md
```diff
@@ -101,21 +101,21 @@ projection.fitSize([width, height], object);
 
 [Source](https://github.com/d3/d3-geo/blob/main/src/projection/index.js) · A convenience method for [*projection*.fitSize](#projection_fitSize) where the width is automatically chosen from the aspect ratio of *object* and the given constraint on *height*.
 
-## Raw projections
+## Raw projections {#raw-projections}
```
mbostock, line 104: This one shouldn’t be necessary since it should be the default anchor for this heading.

mbostock (approved):
Thanks.

I still really want a script we can run against the Markdown at build time so that we can prevent broken links in the future! I think we could do it with just regular expressions if we had to. (If not whatever Markdown parser VitePress uses.)

PR #3667, opened 2023-06-09
stop-gap css for maps in dark mode

Fil:
note: on my computer I get #1e1e1e vs #1e1e20 for the actual background, and it's slightly noticeable. Still better than the white bg!

(next step will be to render them dynamically instead)


mbostock:
I implemented live maps in a4fe425331fceb00b1f65afb472a684a154ecad9. Thanks though!

PR #3669, opened 2023-06-11
add color

Fil:
just an underline; I just saw that this doesn't work in Safari, though?


mbostock:
Use border-bottom: solid 2px color; like we do elsewhere.

mbostock (approved):

PR #3670, opened 2023-06-11
example chord

Fil:
same chord example as a D3 chart instead of an image.

It's not clear to me how to react to dark mode in this example: "brighter" is for dark mode, and should be "darker" for light mode.


mbostock (approved):

mbostock:
Oh, and I dropped the brighter/darker in favor of using the background color for stroking the ribbons.

PR #3673, opened 2023-06-12
test documentation links

Fil:
- Hard errors are expected with 3ae7380 (the script fails to load non-existing files).
- Soft errors are expected with 58c0d2b (some anchors are broken).
- No errors are expected with 3c01a95 (all links fixed).
- I then pushed a version that also supports Plot.


mbostock (approved):
Fantastic!! :clap::clap::clap: I’m so relieved to have this. 😌

PR #3676, opened 2023-06-12
State a general principle for which d3 modules compete with React

tophtucker:
Not sure about the exact phrasing but I think something like this would be nice. It'd be overwhelming to list every module that touches the DOM, but it may be frustratingly vague to just say "most don't, some do". Is there a general principle by which the uninitiated reader might spot which do and don't? As a beginner programmer learning D3 I know I didn't have a good mental model of where the DOM fit in, but I think I'd at least be able to spot "does this take a selection".

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -289,7 +289,7 @@ export default function LinePlot({
 ```
 :::
 
-Some D3 modules (including [d3-selection](./d3-selection.md), [d3-transition](./d3-transition.md), and [d3-axis](./d3-axis.md)) do manipulate the DOM, which competes with React’s virtual DOM. In those cases, you can attach a ref to an element and pass it to D3 in a useEffect hook. (See [sandbox](https://codesandbox.io/s/d3-react-useeffect-5lp0x6?file=/src/LinePlot.jsx).)
+D3 modules that operate on [selections](./d3-selection/selecting.md) (including [d3-selection](./d3-selection.md), [d3-transition](./d3-transition.md), and [d3-axis](./d3-axis.md)) do manipulate the DOM, which competes with React’s virtual DOM. In those cases, you can attach a ref to an element and pass it to D3 in a useEffect hook. (See [sandbox](https://codesandbox.io/s/d3-react-useeffect-5lp0x6?file=/src/LinePlot.jsx).)
 
 :::code-group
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -289,7 +289,7 @@ export default function LinePlot({
 ```
 :::
 
-Some D3 modules (including [d3-selection](./d3-selection.md), [d3-transition](./d3-transition.md), and [d3-axis](./d3-axis.md)) do manipulate the DOM, which competes with React’s virtual DOM. In those cases, you can attach a ref to an element and pass it to D3 in a useEffect hook. (See [sandbox](https://codesandbox.io/s/d3-react-useeffect-5lp0x6?file=/src/LinePlot.jsx).)
+D3 modules that operate on [selections](./d3-selection/selecting.md) (including [d3-selection](./d3-selection.md), [d3-transition](./d3-transition.md), and [d3-axis](./d3-axis.md)) do manipulate the DOM, which competes with React’s virtual DOM. In those cases, you can attach a ref to an element and pass it to D3 in a useEffect hook. (See [sandbox](https://codesandbox.io/s/d3-react-useeffect-5lp0x6?file=/src/LinePlot.jsx).)
```
curran, line 292:
```suggestion
D3 modules that operate on [selections](./d3-selection/selecting.md) (including [d3-selection](./d3-selection.md), [d3-transition](./d3-transition.md), [d3-axis](./d3-axis.md), [d3-drag](./d3-drag.md), [d3-zoom](./d3-zoom.md)) do manipulate the DOM, which competes with React’s virtual DOM. In those cases, you can attach a ref to an element and pass it to D3 in a `useEffect` hook. (See [sandbox](https://codesandbox.io/s/d3-react-useeffect-5lp0x6?file=/src/LinePlot.jsx).)
```

curran (changes requested):
Great idea! I suggested a change to the commit to also mention `d3-drag` and `d3-zoom`, which are also modules that required direct DOM access.

docs/getting-started.md
mbostock: This list isn’t intended to be exhaustive, and I don’t want to set the expectation that we update it if we (theoretically) were to add another module in the future which touched the DOM. It’s just an illustrative list.

mbostock (approved):

PR #3679, opened 2023-06-15
Simplify the quotes for readability

Fil:
If it's ok to use those more as a blurb than as a scientific quote.


mbostock (commented):
I’m not comfortable paraphrasing what other people said and presenting it as a quote. We could say that we’re paraphrasing, but that would nullify the benefit of simplifying the quote. So I say we leave it as is.

Fil:
What is the best service to the reader? Leaner text, or “accuracy”? In this case, I thought the former. In any case, I'd plead for removing the brackets :)

PR #3680, opened 2023-06-15
an intro for axis

Fil:
if we want a more discoverable easter egg, the axis transition could alternatively be triggered on "pointermove click"


mbostock:
I think I would rather show a variety of (3–4?) static axes here—say a time scale, a linear scale, a log scale, an ordinal scale—like we do in the Plot [scales documentation](https://observablehq.com/plot/features/scales#continuous-scales), so that you can quickly see a few different forms of axes at a glance.

Setting tickSizeOuter to zero is also a little surprising since it’s not the default behavior of the axis component. (I’d rather d3-axis didn’t render the domain line by default, and didn’t have outer ticks by default even when it did, but I dunno if we should consider that a breaking change, and at any rate I’m not especially motivated to fix it since we have Plot.)

Maybe we could do a quick transition on pointerenter/pointerleave between two different domains so that we can also use the “static” axes to demonstrate transitions. Or I guess we could do a quick transition on page load; that would be more discoverable and acceptable if it’s ≤750ms.

docs/components/ExampleAnimatedAxis.vue
```diff
@@ -0,0 +1,50 @@
...
+        await transition.end();
+        transition = null;
+      }
+    });
+}
+
+export default {
+  props: {
+    points: {type: Array, default: []},
+  },
```
mbostock, line 39:
```suggestion
```

mbostock:
Given that Plot doesn’t use d3-axis, and that we’re planning on changing Plot to adopt a multi-line tick format for time axes, I think we should write this using D3 to make sure it’s representative of D3’s behavior. I’ll do that now.

mbostock (approved):

Fil:
makes sense! thanks

PR #3750, opened 2023-08-09
Fix d3-time example

kmandov:
- Add missing date arguments to `d3.timeMonth.floor()` and `d3.timeMonth.ceil()` calls.
- Fix the expected date range


mbostock (approved):
Thank you!

PR #3745, opened 2023-07-28
docs: warn about impure `force` functions

hamirmahal:
fixes #3744.

PR #3752, opened 2023-08-13
Revise the force simulation document

kmb645:
The `forceSimulation(nodes)` function operates in an impure manner and can modify its input array of nodes during the simulation.


hamirmahal:
Is this a duplicate of #3745, which was opened 2 weeks before this pull request?

PR #3773, opened 2023-10-05
single module

Fil:
add an example of how to load from a single module (using the recommended ESM method & destructuring)


mbostock (changes requested):
This should be a separate example rather than part of the existing group because it is not functionally equivalent like the other tabs are. I would add some preamble context like “you can also import individual D3 modules like so”.

Other notes:

- Use a major version number when loading d3-force
- Blank lines at the start and end of the script
- No two spaces of indentation for the script
- Use four spaces of indentation for chaining

Fil:
Oops this auto-merged without your approval! I forgot that the rules are different here.

PR #3780, opened 2023-10-11
Edit sentence to fix potential typo and for consistency

sensologica:
I believe that the word should have been "radians", not "radius". Instead of fixing this (potential) typo, however, maybe deleting this part of the sentence altogether is the better option? Deleting it makes the `arc.startAngle(angle)` description match the `arc.endAngle(angle)` description (which does not mention radius/radians in the corresponding sentence). Radians are addressed elsewhere, so mentioning them in the edited sentence is, perhaps, not necessary.


mbostock:
Yes, it was intended to be “radians”.

PR #3782, opened 2023-10-13
always have exactly two/three elements

luizbarboza:


Fil (approved):

docs/d3-scale/diverging.md
```diff
@@ -30,7 +30,7 @@ If *interpolator* is an array, it represents the scale’s three-element output
 const color = d3.scaleDiverging(["blue", "white", "red"]);
 ```
 
-A diverging scale’s domain must be numeric and must contain exactly three values. The default domain is [0, 0.5, 1].
+A diverging scale’s domain must be numeric and must contain exactly three values.
```
mbostock, line 33: Why remove this? It’s correct.

```js
d3.scaleDiverging().domain() // [0, 0.5, 1]
```

mbostock: Ah, it’s already mentioned above. 👍

mbostock (approved):

PR #3795, opened 2023-11-02
document d3-time#66 (consistent intervals with interval.every)

Fil:
Ref. https://github.com/d3/d3-time/pull/66


mbostock:
That PR hasn’t landed yet, so this should be a draft.

PR #3811, opened 2023-12-29
Fix: Typo in README is corrected

codesmith-emmy:


mbostock (commented):
It’s not a typo, but I agree it’s a little awkward. For more than a decade,

- D3 has powered groundbreaking and award-winning visualizations
- D3 has become a foundational building block
- D3 has fostered a vibrant community

Though maybe it should say “has been a foundational building block” rather than “has become a foundational building block”?

codesmith-emmy:
Took a closer look at the sentence again after reading your comment and I have to agree it’s not a typo. This is totally a mistake on my part.

I think “has become a foundational building block” is very much in order and should be left that way.

PR #3816, opened 2024-01-29
Fix svg syntax for JSX code samples in getting-started with React

vimode:
JSX uses `camelCase` convention for props names. 
Updated `stroke-width` accordingly in React code samples.

[React support for SVG components](https://react.dev/reference/react-dom/components#all-svg-components)


mbostock (approved):
Thank you.

PR #3886, opened 2024-05-27
caret dependency ranges

mbostock:
Ref. https://github.com/observablehq/plot/pull/2020

Observable Plot depends on `d3@^7.9.0`, but the semver range is meaningless since D3’s package.json specifies only major versions — you get `d3@7.9.0`, but you might get an older version of `d3-scale-chromatic` that doesn’t have `schemeObservable10`. This ensures that whatever version of `d3` you have, you have at least the latest corresponding transitive dependency version, which makes sense given that `d3` re-exports these transitive dependencies.


Fil (approved):

matthall-akkio:
Do you know when will this be released? Latest on NPM is still 7.9.0 https://www.npmjs.com/package/d3/v/7.9.0

PR #3887, opened 2024-05-28
Clarifying the axis tick's default values

terrence-ou:
Hi, 

when I was first reading this part of the documentation, I felt a bit confused about the second paragraph, in which the text describes the default tick size as 6, but the code still uses the customized tick size and logs out a 0.
original doc

What I did is that I separated the customized tick size declaration from the default one to clarify the difference between the two settings.
modified

Similar modifications were also made in the `axis.tickSizeInner`, `axis.tickSizeOuter`, and `axis.tickPadding` sections.

Thank you for considering this modification!

### DIFF
--- a/docs/d3-axis.md
+++ b/docs/d3-axis.md
@@ -213,12 +213,14 @@ If *format* is not specified, returns the current format function, which default
 
 ```js
 const axis = d3.axisBottom(x).tickSize(0);
+axis.tickSize() // 0
 ```
 
 If *size* is not specified, returns the current inner tick size, which defaults to 6.
 
 ```js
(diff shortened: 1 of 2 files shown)


Fil:
For a less verbose alternative, we could do:

```diff
- axis.tickSize() // 0
+ axis.tickSize() // 0, as specified above
```

terrence-ou:
> For a less verbose alternative, we could do:
> 
> ```diff
> - axis.tickSize() // 0
> + axis.tickSize() // 0, as specified above
> ```

Yeah I agree this is a better solution. Will create another PR for this modification.

PR #3928, opened 2024-10-03
Replace incorrect Wikipedia reference link

maartenverheul:
The [randomBernoulli section](https://d3js.org/d3-random#randomBernoulli) under the `d3-random` docs now links to the Wikipedia page of the `Bernoulli_distribution` instead of the `Binomial_distribution`, which I assumed was incorrect.


Fil (approved):
thanks!

PR #3932, opened 2024-10-10
adds Made by Observable button

CobusT:
This adds the Made by Observable button and new header to the D3 site.


mbostock (commented):
Can we restore the little ↗︎ arrow next to the version number please? Also both these decorations should be in a slightly lighter color as on the current Framework site.

Screenshot 2024-10-10 at 5 34 25 PM

dunstan:
We removed the arrow because all the links we have planned to put in the top right take you somewhere else, as do the links in the "Made by Observable" popup. It felt odd to have arrows on all of them, and odd to have an arrow only on the version number.

Could you share your thoughts on why we'd only have an arrow on that one link, please? And will we also have one next to "observablehq.com" on https://observablehq.com/documentation/ ?

mbostock:
Normally all the links off-site (that open in a new window) have a little ↗︎.

For example the current header on the Plot documentation:

Screenshot 2024-10-10 at 8 40 23 PM

The examples link in the Framework sidebar:

Screenshot 2024-10-10 at 8 41 03 PM

The subdocs in the Documentation sidebar:

Screenshot 2024-10-10 at 8 41 20 PM

This little arrow provides a hint to users that they are leaving the site and that the link will open in a new window.

I violated this rule with the GitHub stars ★ partly as a technical shortcut and more importantly to avoid dual symbols ★↗︎ which felt busy. It was a little bit cheeky to replace the ↗︎ with a ★ to explain what the number meant — sacrificing a bit of utility (since it’s already pretty clear that a link to GitHub will take you off-site).

mbostock:
That said, maybe the arrow isn’t needed because it’s similarly obvious that the release number is an off-site link…

mbostock (approved):

PR #3941, opened 2024-10-22
add rel external to observablehq links

CobusT:
In order for us to track referrals from d3js.org to observablehq.com, we need to add rel = "external" to the links from Vitepress's buttons and navigation.

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -50,7 +50,7 @@ export default defineConfig({
           {text: "What is D3?", link: "/what-is-d3"},
           {text: "Getting started", link: "/getting-started"},
           {text: "API index", link: "/api"},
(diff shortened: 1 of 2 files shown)


mbostock:
Can you point me to the docs for this? I’m curious.

CobusT:
I can't find the official documentation on the VitePress site for this, but [they do mention that](https://vitepress.dev/guide/markdown#external-links), for their Markdown extensions, 'Outbound links automatically get target="_blank" rel="noreferrer"'.

When I looked at the links generated by VitePress for the links addressed in this PR, it added `rel="noreferrer"` as a default, it seems. Explicitly specifying `rel="external"` fixes this.

mbostock:
I see, so it’s just that we’re overriding the `rel` attribute to not have the default `noreferrer`. 👍

mbostock (approved):

PR #3952, opened 2024-11-21
docs: describe property as optional in d3-format locale

djbarnwal:
This PR updates the docs for the package `d3-format`

As per the [source code](https://github.com/d3/d3-format/blob/f3cb31091df80a08f25afd4a7af2dcb3a6cd5eef/src/locale.js#L17) and [example usage](https://d3js.org/d3-format#formatLocale), `decimal` is also an optional attribute for the locale config. This PR updates the docs to reflect that.


lbesecker195 (approved):
Verified against `d3-format/src/locale.js` line 17: `decimal = locale.decimal === undefined ? "." : locale.decimal + ""`, so it is optional with a `.` default. The reorder groups it with the other optional fields. LGTM.

PR #4102, opened 2026-03-30
docs: fix typos in changelog

ayushshukla1807:
Fix minor typos in changelog. Found via codespell.


Fil:
Thanks. But this is a changelog from 2016, I don't think we need to fix minor typos.

